Issue #56, opened 2019-12-27
Site menu overhaul


leecalcote:
**Current Behavior**
The current menu has "GitHub" listed, but is missing Slack, the mailing list, community drive, community calendar and so on.

**Desired Behavior**
New top-level menu items with sub-items structured like so:

[Docs](https://meshery.layer5.io/docs)

[Community](/community)

[Blog](https://layer5.io/blog)

Connect
- [Calendar](https://layer5.io/subscribe/)
- [GitHub](https://github.com/layer5io/meshery)
- [Mailing List](https://layer5.io/subscribe/)
- [Slack](http://slack.layer5.io)

Resources
- [Branding](https://layer5.io/brand#meshery)
- [Code of Conduct](https://github.com/layer5io/meshery/blob/master/CODE_OF_CONDUCT.md)
- [Contributing](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md)

[Run Meshery](#getting-started)

ashiscs:
@leecalcote sir, can you elaborate this issue?

leecalcote:
@ashiscs good question. I've updated the description of the issue. Does this help?

l5io:
Let's be consistent and reuse the same menu javascript/html from layer5.io here on meshery.io.

leecalcote:
Thank you for triaging issues, @ashiscs 👍

PR #71, opened 2020-01-17
#56 meshery.io new nevbar options are added

gunjanmimo:
Signed-off-by: gunjanmimo <gunjan.mtbpaul@gmail.com>


ashiscs:
fixes #56

leecalcote:
Thank you, gents. Let's close this one in lieu of #81 (@supra08). There's are a couple issues that need to be addressed, yet - some in mobile view and one when the top-level menu item is clicked.

PR #81, opened 2020-01-22
Navbar Overhaul

supra08:
In this pull request, I've implemented the new navbar with submenus. Currently, it works fine in the desktop view. Some bugs still do exist in the mobile view mainly because of stellarnav's inherent styling.

@leecalcote


ashiscs:
fixes #56

Issue #82, opened 2020-01-23
Mobile View: Navigation menu


leecalcote:
**Bugs to Fix**
1. Font should be white.
1. Menu should be in upper right hand corner.
1. Top-level navigation items should not have a hyperlink (should not reload the page).

**Screenshots**
<img width="513" alt="Screen Shot 2020-01-23 at 8 12 58 AM" src="https://user-images.githubusercontent.com/7570704/72991847-59515280-3db8-11ea-855c-9567757f4cff.png">

leecalcote:
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/7570704/72992859-109a9900-3dba-11ea-9b4a-ee4ef595daea.gif)

leecalcote:
The new menu is great, @supra08! 👍 

There are couple residuals issues to tidy up.

Nikhil-Ladha:
Here , I think only the last part of the bug is left out, as the first first bug was fixed in my previous PR itself. Correct @leecalcote ?

leecalcote:
Yes, I think the only residual issue here is that the hamburger (mobile) menu is left aligned, not positioned in the upper, right hand corner.

Nikhil-Ladha:
> Yes, I think the only residual issue here is that the hamburger (mobile) menu is left aligned, not positioned in the upper, right hand corner.

I meant the last issues of page reload on top-level navigation items. The hamburger has been right aligned in the changes made in the #237 PR. You can confirm it from the meshery.io site.

PR #85, opened 2020-01-25
Set mobile menu to white

leecalcote:
- Changed mobile hamburger menu to white color.
- Added distinction between dev and prod environments for legibility of css and js for stellarnav.


leecalcote:
This is a partial fix for #82

PR #87, opened 2020-02-04
Feature/helm chart

Aisuko:
The Meshery chart repository for Helm Hub.


Aisuko:
> @Aisuko the changes include here remove all files that currently comprise the Meshery project site (https://meshery.io). Can you refactor to leave existing files in-place and use `/charts` folder to host `index.yaml`, which will point to the charts hosted in the `meshery` repo?

Done, please kindly have re-check @leecalcote Apologize for I did not understand your mean at the first time. But I'm not sure the chart's repository will work. I mean it's versus to docs may got complex. So, please after the PR merge. We need to check Github page.

PR #88, opened 2020-02-08
Add deploy meshery by Helm chart. The project's ruby dependency version

Aisuko:
Signed-off-by: Aisuko <urakiny@gmail.com>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+ruby '2.6.5'
+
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+ruby '2.6.5'
```
Aisuko: The version dependency is the same as the [layer.io](https://github.com/layer5io/layer5).

leecalcote: Good call.

PR #90, opened 2020-02-09
Init the Helm chart

Aisuko:
Signed-off-by: Aisuko <urakiny@gmail.com>

As we talk, we use the master branch. Please do not forget to delete the feature/helm-chart branch.


charts/_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-slate
```
leecalcote: I don't know that this config will be used. We might want to consider its removal, if it isn't.

charts/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
leecalcote: Will this LICENSE file be used as part of the Helm Hub entry? If not, the LICENSE file found in the root of this repo might suffice unto its own.

charts/index.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - meshery
+    - mesheryistio
+    - mesherylinkerd
+    maintainers:
+    - email: urakiny@gmail.com
+      name: aisuko
+    name: meshery
+    type: application
+    urls:
+    - meshery-0.1.0.tgz
```
leecalcote: Given that the source for these charts in is the Meshery repo, instead of storing the compressed version of the charts in this repo as a charts "release", can we consider incorporating the tar.gz creation of these charts to our GitHub Actions workflow in Meshery? 

We would include the tar.gz creation part of the tagged workflow (the workflow that is enacted when a release tag is present). We could either have goreleaser tar.gz the charts or have a script executed to tar.gz the charts and make that tar.gz available alongside the rest of the Meshery release artifacts here - https://github.com/layer5io/meshery/releases

In this way, the Meshery helm chart is treated as another release artifact that is created each time we cut a new Meshery release.

leecalcote:
Branch `feature/helm-chart` deleted.

charts/_config.yml
Aisuko: The theme used for the chart web UI.

charts/LICENSE
Aisuko: Yes, I believe we can remove this.

charts/index.yaml
Aisuko: Good idea, maybe we can do this. The Meshery image's version was used `latest` under the chart. This tag which you said is the chart itself version. The Meshery version in the chart is here below.
 https://github.com/layer5io/meshery/blob/8596379830f5826379923afb4e24890b655f52b2/install/kubernetes/helm/meshery/Chart.yaml#L34

charts/_config.yml
Aisuko: I think we can remove this.

PR #111, opened 2020-02-27
external navigation links

leecalcote:
Adding new icons to denote that a navigation item links to an external website.


leecalcote:
Thanks, @arghac14

Issue #127, opened 2020-03-18
Incorporate display of " xxx performance tests run!"


leecalcote:
**Current Behavior**
Meshery users send performance test results to the Meshery Cloud provider, but the total number of results collected isn't displayed on the site.

**Desired Behavior**
Similar to how https://layer5.io/meshery uses a public endpoint to display how many performance tests have been run, let's find a spot on meshery.io to display and explain this statistic.

<img width="655" alt="Screen Shot 2020-03-18 at 2 55 56 PM" src="https://user-images.githubusercontent.com/7570704/77001840-97867080-6928-11ea-98fc-8aa2d20b25a8.png">

gunjanmimo:
![Screenshot from 2020-03-24 01-34-17](https://user-images.githubusercontent.com/46948185/77358307-9eb9de00-6d6f-11ea-8c4b-ffc7f07d5ad7.png)


@leecalcote sir, is this right place to show the performance test of Meshery?

kanishkarj:
@gunjanmimo @leecalcote , I think that fits in perfectly with the rest of the UI! I Think, we could go ahead with this.

leecalcote:
I agree with @kanishkarj. Double check, if you would, @gunjanmimo, is there space that we might eliminate between the current number and the rest of the "performance tests run!" sentence?

PR #132, opened 2020-03-23
Performance test gunjanmimo 127

gunjanmimo:


ashiscs:
fixes #127

_includes/getting-started.html
```diff
@@ -1,22 +1,111 @@
-<div class="anchor"><a name="getting-started"></a><h2>Run Meshery</h2></div>
-<div class="container flex">
+<html>
+  <head>
+
```
leecalcote: Given that this is a partial html snippet that will be shown in context of a larger page, we'll want to remove the `<html>` and <head> tags.

_includes/getting-started.html
```diff
@@ -1,22 +1,111 @@
...
+          width: 100%;
+          position: relative;
+          top: 300px;
+          left: -630px;
+        }
+
+      }
+
+     
+    </style>
```
leecalcote: It would be good to pull there into an existing or new sass file - https://github.com/layer5io/meshery.io/tree/master/_sass

gunjanmimo: @leecalcote sir, this style section for getting started section?

leecalcote: @gunjanmimo right, yes. This document `getting-started.html` is a partial html snippet. If we can, it's best to land these new styles into an existing (or create a new) sass document.

gunjanmimo: @leecalcote  sir, i have resolved this issue and added all the styling in respective sass file

PR #143, opened 2020-03-31
fix: Only use blue-colored text when text is hyperlinked

### LINKED ISSUE #142: [Getting Started] Only use blue-colored text when text is hyperlinked
**Description**
Blue-colored text is used on this site to indicate to the user that the text is hyperlinked. Do not use blue text without a hyperlink.

**Expected Behavior**
Text without a hyperlink should be white in color.

**Screenshots**
Screen Shot 2020-03-30 at 9 36 47 PM

kushthedude:
fixes #142


kushthedude:
Done

PR #150, opened 2020-04-02
Addition of rasterized service mesh logos

leecalcote:
This PR simply adds rasterized versions of some service mesh logos, where we already have vector-based copies of the same logos. GH markdown doesn't want to display SVGs, so these PNGs are necessary for displaying in READMEs.


leecalcote:
Thanks, gents.

PR #154, opened 2020-04-02
correct the build command

kushthedude:
Build command was not correct as there is no `--live-reload` flag with build option


leecalcote:
How long does an individual PR preview stick around?

kushthedude:
Forever I suppose, I can still see PR deploy previews of my commit which
are 2.5 years old.

PR #242, opened 2020-07-01
Added Meshery Presentation

### LINKED ISSUE #172: Publish Meshery Presentation
**Current Behavior**
The [Meshery - The Service Mesh Management Plane](https://drive.google.com/open?id=14kxjwYSJ_FyE3K_6CDEd6oq2kqwn0OSE8RDJ4H-KlKU) presentation gives an overview of the project and related projects. 

**Desired Behavior**
It would be good to have on the meshery.io site as a published Google Slide.

Akanksha1212:
Fixes #172 
Signed-off-by: Akanksha1212 <akankshasingh121297@gmail.com>


Nikhil-Ladha:
I think we would prefer to give a heading too for the slides section something like 'Meshery Presentation' itself.

NupurThakur27:
> I think we would prefer to give a heading too for the slides section something like 'Meshery Presentation' itself.

Don't you think that would be a redundant information?

Nikhil-Ladha:
> > I think we would prefer to give a heading too for the slides section something like 'Meshery Presentation' itself.
> 
> Don't you think that would be a redundant information?

We may change the heading but having a slide show in the middle without any context seems a bit abrupt to me. If everyone agrees with staying this way, then I am ok with it too. Just a thought.

NupurThakur27:
> > > I think we would prefer to give a heading too for the slides section something like 'Meshery Presentation' itself.
> > 
> > 
> > Don't you think that would be a redundant information?
> 
> We may change the heading but having a slide show in the middle without any context seems a bit abrupt to me. If everyone agrees with staying this way, then I am ok with it too.

Your point seem  fair too, let's see what everyone have to say about it.

index.html
```diff
@@ -193,6 +193,17 @@ <h2>Which service mesh should I use and how do I <strong>get started</strong>?</
 			
 		</div>
 	</section>
+
+	<section>
+		<div class="container">
+			<div class="google-slides" >
+				<iframe src="https://docs.google.com/presentation/d/e/2PACX-1vSJdaOJ6MdmYQCyd2c3sbGsa9h7BVqauzBpdeVIKMpAh4eHoStLTMS8hmYJdzT0YTgXzS2chyrLPRzR/embed?start=true&loop=true&delayms=5000" 
+				frameborder="0" allowfullscreen="true" mozallowfullscreen="true" webkitallowfullscreen="true"></iframe>
+			</iframe>
```
ashiscs, line 202: Extra </iframe> tag can be removed !

leecalcote:
We might do well to include the presentation into its own section (with a distinct background color - perhaps, that of the footer). A fresh look after the addition of a separate section might give a different perspective on whether a section header is or is not an improvement.

We'll want to ensure that the ratio for the iframe size is precise such that it doesn't leave any empty (black) space.

Screen Shot 2020-07-02 at 12 00 13 AM

Akanksha1212:
Thanks @leecalcote for the feedback. Will make changes and add them.

index.html
Akanksha1212: Okay. Will remove that

leecalcote:
@Akanksha1212 I tossed the presentation into a new section with a heading. What do you think?

Nikhil-Ladha:
@leecalcote there is an alignment issue with the heading 'Overview', if you can fix that. It shall look better.
Screenshot_20200708-000950

leecalcote:
@Nikhil-Ladha sounds good. I don't know that I see the alignment issue. Is the "Overview" word to far to the left/right or up/down?

It is right-aligned to stop at the same width of the google slides iframe. The small difference between the actual slide and the black background in the google slide iframe may be deceiving in terms of alignment.

Nikhil-Ladha:
> @Nikhil-Ladha sounds good. I don't know that I see the alignment issue. Is the "Overview" word to far to the left/right or up/down?
> 
> It is right-aligned to stop at the same width of the google slides iframe. The small difference between the actual slide and the black background in the google slide iframe may be deceiving in terms of alignment.

I think the word should be left aligned like other headings on the page, to keep consistency.No issue with the small black background in the iframe.

Akanksha1212:
Hi @leecalcote, It looks good on a smaller screen but I think some part of the slide is not visible on bigger screen size.
image

Nikhil-Ladha:
> Hi @leecalcote, It looks good on a smaller screen but I think some part of the slide is not visible on bigger screen size.
> image

Good point, I guess the height of the iframe should be reduced a bit.

leecalcote:
Interesting. Curious-what’s the resolution of your display?

@Akanksha1212, if you would like to pick up from here and make changes, please do. For the most part, I jumped in to give you some encouragement and let you know that, for my part, I am eagerly awaiting this update. 😀

Akanksha1212:
It's 1920*1080. I'll try to fix the height.

PR #257, opened 2020-07-07
update quick start instructions

### LINKED ISSUE #256: Update Quick Start instructions
**Description**
The second quick start instruction is wrong: "Access Meshery in your browser if you are using Kubernetes at http://127.0.0.1:9081."

**Expected Behavior**
The instruction should be "Open Meshery in your browser."

morrme:
Fixes #256 

Signed-off-by: morrme <morrme@users.noreply.github.com>

### DIFF
--- a/_includes/getting-started.html
+++ b/_includes/getting-started.html
@@ -3,7 +3,7 @@
   <h1 class="getting-started-heading">Getting Started</h1>
     <h2 style>with Meshery</h2>
       <p style="font-size: 1.15rem;">
-        Choose a platform and see how easy it is to get up and running with a service mesh.
+        Choose a platform and see how easy it is to start managing service meshes in a single command.
       </p>
     </h1>
 </div>
@@ -13,7 +13,7 @@
     <div class="post-content row" itemprop="articleBody">
       <div style="display: inline-flex; line-height: 60px;">
         <h1 class="step-number">01</h1>
-        <h2 class="step-heading">Install and Start</h2>
+        <h2 class="step-heading">Choose your Platform</h2>
       </div>
       <div class="platformIcon">
         <label class="tooltip" for="start-helm-v2" style="margin: 0 20px 0px 0px; padding: 0;">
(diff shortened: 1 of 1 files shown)


morrme:
preview

Nikhil-Ladha:
> Thanks for this screen shot. These are a timesaver for reviewers. 
> 
> It's nice to see what the Open Sans font looks like on different systems.

Shouldn't we keep the address to point out to after configuring Meshery. It shall be helpful for users.

chandrashritii:
@morrme Good job.

@leecalcote I agree with @Nikhil-Ladha.
Mentioning the address would be handy for  users, especially beginners.

leecalcote:
You all are right. Presenting the user with URL is a better experience.

Consider though that that URL is specific to Docker deployments only.

Nikhil-Ladha:
Oh okay, then we can consider mentioning that. Another thing, what's the case then for other deployments then?

morrme:
@leecalcote @Nikhil-Ladha @ashiscs @chandrashritii  Thank you for the input!
Should I hold off until a consensus is reached ?

leecalcote:
> Oh okay, then we can consider mentioning that. Another thing, what's the case then for other deployments then?

@Nikhil-Ladha great suggestion. The issue is that those other cases, particularly Kubernetes, is long and depends on any number of factors. We won't want to list those here, but point to the docs.

leecalcote:
@morrme thanks for asking. I should've used more prescriptive language above. :)

In the default Docker deployment, Meshery will automatically open their browsers to http://localhost:9081. So, we don't even need to list a second step. It's a one shot motion from download, to install, to start. It's pretty sweet and we should market this aspect of the project better.

The Kubernetes deployments are varied and will take much verbiage to explain, so we won't. The Homebrew and Windows deployments have about three steps total, so they're pretty quick, but not something that we want to prominently list.

And so, while my initial ask was to reduce line 2 to a shorter phrase ("Open Meshery in your browser."), I'm going to update the request to change the section to say:

"Getting Started
with Meshery

Choose a platform and see how easy it is to start managing service meshes in a single command.

01 Choose your Platform

02 Manage
   There is no step 2. Login and manage service meshes!"

Nikhil-Ladha:
> > Oh okay, then we can consider mentioning that. Another thing, what's the case then for other deployments then?
> 
> @Nikhil-Ladha great suggestion. The issue is that those other cases, particularly Kubernetes, is long and depends on any number of factors. We won't want to list those here, but point to the docs.

Gotcha!

PR #259, opened 2020-07-09
Fixed navigation menu

### LINKED ISSUE #82: Mobile View: Navigation menu

Nikhil-Ladha:
This PR fixes #82 
Signed-off-by: Nikhil-Ladha <nikhilladha1999@gmail.com>

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -12,6 +12,7 @@
   new_window: false
   highlight: false
 - name: Connect
+  submenu: true
   new_window: false
   highlight: false
   subitems:
@@ -36,6 +37,7 @@
       new_window: true
       external: true
 - name: Resources
+  submenu: true
   highlight: false
   subitems:
     - name: Branding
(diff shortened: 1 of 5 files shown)


Nikhil-Ladha:
> The line breaks here look a bit worse than what we have currently:
> 
> Screen Shot 2020-07-09 at 12 03 48 PM

Well, for this I have to use a hack. No other option can be thought of there. Hack can be setting a width of 125px to the resources option only or all of them to compensate the width of all the options in one line

Nikhil-Ladha:
But anyways, all are of different width so, applying a width to only the `resources` option doesn't looks like much of a harm to me.

Nikhil-Ladha:
@leecalcote shall we go with this?

PR #260, opened 2020-07-09
Refactored community resource links

leecalcote:


leecalcote:
@morrme you may want to provide review on this. If you find that you can't, you should have an invite to join the layer5io org here - https://github.com/layer5io.

morrme:
@leecalcote for consistency, there should be an agreement on whether or not periods should be placed inside or outside of the closing </a> tag. 

links

leecalcote:
Good call, @morrme. Period moved to outside the anchor tag.

PR #265, opened 2020-07-17
Added gemfile.lock

crustyy:


kushthedude:
Gemfile.lock should be in your repository. It's used as a snapshot of all your required gems and their dependencies. This way bundler doesn't have to recalculate all the gem dependencies each time you deploy, etc. 
To maintain a consistent version, to improve the bug tracking & dependency management I suggest we should a gemfile.lock & move ahead with the PR.

PS:- If anyone still has some doubt about the decision, here's a nice article explaining the same https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

PR #327, opened 2020-09-22
corrected the navbar issue

reveurguy:
**Description**
corrected the navbar color issue.
This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/_sass/blog.scss
+++ b/_sass/blog.scss
@@ -1,3 +1,6 @@
+header {
+	background-color: $coolgray;
+}
 .blog-posts {
 	list-style: none;
 	padding: 0;
--- a/_sass/features.scss
+++ b/_sass/features.scss
@@ -1,3 +1,6 @@
+header {
+	background-color: $coolgray;
+}
 .features {
 	@extend %flexbox;
 	@include flex-flow(wrap);
--- a/_sass/layout.scss
+++ b/_sass/layout.scss
@@ -29,7 +29,7 @@
 }
 
 header {
-	background-color: $coolgray;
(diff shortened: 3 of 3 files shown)


Neilblaze:
@reveurguy good job!

But I would like to point out two things over here.

- The navbar of the [meshery](https://meshery.io/) site doesn't look like what it is in your deploy [preview](https://deploy-preview-327--mesheryio.netlify.app/). Notice that in the [Meshery](https://meshery.io/) landing page & [programs](https://meshery.io/programs/) page, the navbar seems to be blended into the background with a greenish gradient (follow [Layer5 brand](https://layer5.io/brand) guide for more info). What you have done is setting the background color of the navbars to `#3C494F`, which does fix the problem for #326 & #325 , but at the same time reflects change on both the pages. So, try to tackle it differently!

- Fix the DCO. Follow [this](https://github.com/layer5io/layer5/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin) if you get stuck!

reveurguy:
@Neilblaze is there a single header for all pages? or different headers for different pages?

Neilblaze:
In jekyll, to create a full fledged site.page, we break the page into different segments & each of the segments have their own css file. So for different pages there will be different css files or .sass file which post converts to the root css file. A small piece of advice from my side, try to avoid inline css over here!

reveurguy:
@Neilblaze i get this error when i am doing git commit -m -s "" :  pathspec " " did not match any files known to git.

PR #377, opened 2020-11-12
added envoy filter section

### LINKED ISSUE #350: Add EnvoyFilter Provisioning as a feature to the documentation
**Desired State:**

We have introduced a new feature for meshery to configure a custom EnvoyFilter of their choice for the Istio service mesh. We should have them added to the main-page as well as the `/features` section.
Here is a sample screenshot that can be attached to describe the feature on meshery dashboard.
Screenshot 2020-10-16 at 6 44 40 PM

---
**Contributor Resources**
- [Meshery documentation site](https://meshery.layer5.io/docs/)
- [Meshery documentation source](https://github.com/layer5io/meshery/tree/master/docs)
- [Instructions for contributing to documentation](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md#documentation-contribution-flow)

kumarabd:
Signed-off-by: kumarabd <abishekkumar92@gmail.com>

**Description**

This PR fixes #350

**Notes for Reviewers**

- [ x] Yes, I signed my commits.


Nikhil-Ladha (commented):
Everything looks good, just one thing you should not commit the lock file. Please undo that.

kumarabd:
@Nikhil-Ladha Lets not merge before I upload the assets

Nikhil-Ladha:
> @Nikhil-Ladha Lets not merge before I upload the assets

Cool, you can mark is as draft for a safer side.

kushthedude:
> Everything looks good, just one thing you should not commit the lock file. Please undo that.

Why so? Lock file should be committed as is already present in the repo for consistent versioning of ruby deps.

kushthedude (changes requested):
> Everything looks good, just one thing you should not commit the lock file. Please undo that.

Why so? Lock file should be committed as is already present in the repo for consistent versioning of ruby deps.

Nikhil-Ladha:
Lock file might cause issues in some systems if the versions of deps differ, then they will have to remove it and build it again. That's the reason I think it's better not to change it unless we have some changes in the Gemfile itself.

kushthedude:
So lockfile is deleted right now, that is what you want?

Nikhil-Ladha:
I think that would be better, but if you think the other way please do share that too. I would love to ve corrected regarding that so that we follow this our other repos too.🙂

kushthedude:
> I think that would be better, but if you think the other way please do share that too. I would love to ve corrected regarding that so that we follow this our other repos too.🙂

Sure please https://github.com/layer5io/meshery.io/pull/265#issuecomment-660287282 , and I guess this is what is followed in other repos if not then should be.

Nikhil-Ladha:
So, in that case let's revert the changes and keep it as it was till now. By which I mean let's update the lock file with the changes.

PR #419, opened 2021-04-01
Add Dynamic SMI Table

utkarsh-pro:
**Description**

This PR adds dynamic SMI Table support on `https://meshery.io/smi`. Jekyll doesn't supports dynamic content out of the box so a new plugin which adds a new fluid tag to jekyll is also added here.

Usage of the newly created tag: `{% externalJSON data from url <api-endpoint-here> %}`

Relevant Screenshot:
Screenshot_20210401_175323

**Notes for Reviewers**

### DIFF
--- a/_plugins/externaljson.rb
+++ b/_plugins/externaljson.rb
@@ -1,0 +1,26 @@
+require 'json'
+require 'net/http'
+
+module ExternalJSON
+  class ExternalJSON_tag < Liquid::Tag
+
+    def initialize(tag_name, text, tokens)
+      super
+      @text = text
+    end
+
+    def render(context)
+        if /(.+) from url (.+)/.match(@text)
+          resp = Net::HTTP.get_response(URI($2.strip))
+          data = resp.body
+          context[$1] = JSON data
+          nil
+        else
+          # syntax error
+          raise ArgumentError, 'ERROR:bad_syntax'
+        end
+    end
+  end
+end
+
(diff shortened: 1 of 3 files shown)


leecalcote:
With smi-spec.io running on Hugo, I suppose this means that we'll need an alternative implementation.

Using the `externalJSON` plugin, the table data is refreshed on page load, correct?y

utkarsh-pro:
> With smi-spec.io running on Hugo, I suppose this means that we'll need an alternative implementation.
> 
> Using the `externalJSON` plugin, the table data is refreshed on page load, correct?

Yes, using the plugin the data is fetched before rendering the page on the server side and then fed to jekyll (hence jekyll consider it another static json).

I am not sure if I understood the alternative implementation part :sweat_smile:

dhruv0000 (approved):

leecalcote:
@dhruv0000 we're missing the version of the SMI Conformance test (the Learn Layer5 release) in this table. Is that version available in the results endpoint?

leecalcote:
@utkarsh-pro whee! The "alternative implementation" bit is in reference to the next step in this effort being the publishing this same table of results on https://smi-spec.io, which runs on Hugo.

dhruv0000:
> @dhruv0000 we're missing the version of the SMI Conformance test (the Learn Layer5 release) in this table. Is that version available in the results endpoint?

Yes it is. Let me fix this.

Issue #429, opened 2021-04-18
Programs page is covered by footer 


Kn99HN:
**Description**
Programs page at `https://meshery.io/programs` is hidden by the footer

**Expected Behavior**
A brief description of what you expected to happen.

**Screenshots**
Add screenshots, if applicable, to help explain your problem.
<img width="1440" alt="Screen Shot 2021-04-17 at 5 13 09 PM" src="https://user-images.githubusercontent.com/44149581/115126876-4f5cd600-9fa0-11eb-9cba-b01e41e20c99.png">

**Enviroment:**
 - OS: MacOS
 - Browser: Firefox
 - Version: 88.0b9
 - Device: Macbook Pro

---
[Optional] **To Reproduce**
1. Go to 'https://meshery.io/programs/'
3. Scroll down to Subcribe form

[Optional] **Additional Context**

Josephchinedu:
In the `https://meshery.io/css/screen.css` add `margin-top: 300px;` to the footer styling

leecalcote:
Good call, @Josephchinedu. Would you like to take this issue?

leecalcote:
Thank you, @Josephchinedu 👍

leecalcote:
Be sure to reference this issue in your PR's description (`This PR fixes #`) so that this issue will automatically be closed out.

ishitakeshawani:
@Josephchinedu are you working on it?
Otherwise I like to solve it @Nikhil-Ladha

PR #430, opened 2021-04-22
Programs page is covered by footer #429: issue resolved

Josephchinedu:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/_sass/footer.scss
+++ b/_sass/footer.scss
@@ -104,6 +104,7 @@ footer {
 	position: relative;
 	background: #222;
 	color: #fff;
+	margin-top: 300px;
 
 	.copyright {
 		font-size: .8em;
--- a/css/program.css
+++ b/css/program.css
@@ -85,7 +85,7 @@ a:hover {
 
 a:active {
   text-decoration: none;
-}k
+}
 .appear {
   display: none;
 }
--- a/css/program.css
+++ b/css/program.css
@@ -85,8 +85,7 @@ a:hover {
 
 a:active {
   text-decoration: none;
-}
-
+}k
 .appear {
   display: none;
 }


Nikhil-Ladha (changes requested):
You should not be changing the footer CSS directly, it impacts all that pages.
Instead look at the component, which has the issue and the fix should be restricted to that only.

Also, please mention the issue number in the description `This PR fixes #<issue-number>` it will automatically close the issue when the PR is merged.

leecalcote:
@Josephchinedu thanks for working on this one. Will you take @Nikhil-Ladha's suggestions into consideration?

Josephchinedu:
@leecalcote i actually tried to implement meaning @Nikhil-Ladha suggestion, but i don't know to do that without touching the footer CSS. and i know it'll affect other pages.

Nikhil-Ladha:
> @leecalcote i actually tried to implement meaning @Nikhil-Ladha suggestion, but i don't know to do that without touching the footer CSS. and i know it'll affect other pages.

Firstly, you need to rebase your local fork with the latest upstream code. In the current preview, the mailing list div is not even there.
After doing so, the only fix that you need to do is remove this CSS:->
https://github.com/layer5io/meshery.io/blob/master/css/program.css#L119

leecalcote:
Any suggestion on how to proceed?

Nikhil-Ladha:
> Any suggestion on how to proceed?

The solution is being highlighted in the previous comment, now waiting for the update from @Josephchinedu

NikhilSharma03:
@adithyaakrishna I think this one is also related to the footer problem. Can you please check?

Jashpatel1:
The issue has already been solved by another contributor.

PR #650, opened 2021-12-27
Changes in video modal

shubham-tyagi:
* Made the video modal responsive, 
* Made the fullscreen button work for ESC key, 
* Added the functionality to stop or play the video from the space bar
* If the modal is opened then the page remains still till we don't close the modal.

Signed-off-by: shubham-tyagi <raj.tyagi2000@gmail.com>

**Description**

This PR fixes #

**Notes for Reviewers**
Can you check if now the video modal is working fine in safari or not? 

### DIFF
--- a/_includes/partials/patterns.html
+++ b/_includes/partials/patterns.html
@@ -19,7 +19,7 @@
                 <a class="link recording-link">See service mesh patterns in-action</a>
                 <p style="font-size: 1rem;color:#666;font-style: italic;">(video excerpt)</p>
             </div>
-            <div id="open-modal" class="modal-window">
+            <div id="open-modal" class="video-modal">
                 <div class ="modal-content">
                   <button class = "modal-close close">&times;</button>
                   <h1>Service Mesh Patterns in-action</h1>
--- a/_sass/patterns.scss
+++ b/_sass/patterns.scss
@@ -1,361 +1,416 @@
 
-  /* MODAL-STYLES */
-  .modal-window {
-    position: fixed;
-    background-color: rgba(0, 0, 0, 0.75);
-    top: 0;
-    right: 0;
-    bottom: 0;
-    left: 0;
-    z-index: 999;
-    visibility: hidden;
-    opacity: 0;
-    pointer-events: none;
-    transition: all 0.3s;
-  }
-  .modal-image {
-    width: 250px;
-      margin: 30px 30px 2rem 0px;
-      float: left;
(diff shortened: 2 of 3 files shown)


warunicorn19:
It works better now in safari, wouldn't say it is 100% compatible yet, but we can work with that. Just one thing, I feel the default size of the modal is a bit too small. Could you please make it a bit bigger?
Screenshot 2021-12-27 at 10 58 43 AM

shubham-tyagi:
yeah sure @warunicorn19, on it.

leecalcote:
I was just thinking about this an hour ago.

leecalcote:
Any way to be rid of the empty black space (have the frame fit the video)?
Screen Shot 2021-12-31 at 11 14 04 AM

leecalcote:
@shubham-tyagi this approach seems convoluted and unnecessarily complex. Can you go back to the drawing board and reassess this (entire) approach?

What are the two independent timelines for?
Screen Shot 2021-12-31 at 11 44 48 AM

leecalcote:
@shubham-tyagi thank you for digging into this. Will you review https://github.com/meshery/meshery.io/pull/651 in consideration of an alternate implementation?

shubham-tyagi:
@leecalcote the second timeline is not a timeline, it is basically used to adjust the volume of the video through the video player.

shubham-tyagi:
Yeah, actually this approach is fairly complex. I was thinking of having a bit more customized video player with some nice hover effect. Though it's not necessary, I thought that would look nice. I thought of using a different video player API as well but didn't find a simple one that I could use easily so I went with the traditional JS approach.

shubham-tyagi:
@leecalcote I could maybe even add some useful comments in the script to explain the code and make it fairly simple to update the script later and have an include file for the video modal like you did in your PR, what do you think about it ?

leecalcote:
Thanks for suggesting this. It would be redundant effort. We have hundreds of thousands of lines of code to maintain across the projects. We're just looking for something simple here.

shubham-tyagi:
@leecalcote it does make sense, will try and keep the code as simple as possible going forward.

leecalcote:
You did a fine job here. The custom player is neat. It gives the site a special feel. 

Simplicity is something that we need to strive for more in some of the other projects, in fact.

leecalcote:
Let's close this one out so that you can focus on a few of the others that you have assigned.

Happy New Year! 🎉

PR #687, opened 2022-03-28
[Setup]Fixed Errors at setting up website in Windows OS

Harshit-Dandriyal:
Signed-off-by: Harshit-Dandriyal <dandriyalh9318@gmail.com>

**Description**
The {Make site} is working fine in ubantu without this PR

 
ubantumakesite

But, In Windows It was creating a error 

 
Screenshot 2022-03-29 033838

Now, On many Unix-based systems (e.g. Linux), TZInfo is able to use the system [zoneinfo] directory as a source of data. However, Windows doesn't include such a directory, so the tzinfo-data gem needs to be installed instead
but still After adding 
gem 'tzinfo-data'
 gem 'tzinfo'
 in Gemfile
Another Error occurs

differror
 
It was solved by adding
gem 'eventmachine', '1.2.7', git: 'https://github.com/eventmachine/eventmachine.git', tag: 'v1.2.7'
in gemfile

This PR fixes #
[Setup]Fixed Errors at setting up website in Windows OS

**Notes for Reviewers**


leecalcote:
@cpepper96 will you have a look this PR?

cpepper96 (commented):
Yeah this worked on my local machine but interestingly I ran into the same redirect loop that first popped up with running just the `bundle exec jekyll serve` command. Maybe it's not recognizing the config file somehow?

cpepper96:
I just realized this is not meshery/meshery repo. Whoops, I will have to look at this tomorrow 😂

cpepper96:
yep this works on my machine. @Harshit-Dandriyal what is the .eventmachine.swp file? Is that necessary? It works on mine without it. 
CC: @leecalcote

Harshit-Dandriyal:
@cpepper96 I already sent my list of error coming up in this PR. Also, It's happening in windows platform only.

Harshit-Dandriyal:
@cpepper96 I think that tz-info dependecy error will be coming up in every windows platform

leecalcote (changes requested):
@Harshit-Dandriyal please remove the .swp file from this PR.

Harshit-Dandriyal:
@leecalcote Done 👍

leecalcote:
@Harshit-Dandriyal, what are the eventmachine and tzinfo_data packages used for? Why are they necessary for building this site on Windows and not on Mac and Linux?

Harshit-Dandriyal:
@leecalcote 
The TZInfo::DataSourceNotFound error is being raised by TZInfo, a dependency of the Active Support component of Rails. TZInfo is looking for a source of time zone data on your system, but failing to find one.

On many Unix-based systems (e.g. Linux), TZInfo is able to use the system [zoneinfo](http://www.iana.org/time-zones) directory as a source of data. However, Windows doesn't include such a directory, so the tzinfo-data gem needs to be installed instead. The tzinfo-data gem contains the same zoneinfo data, packaged as a set of Ruby modules.

Rails generates a default Gemfile when the application is first created. If the application is created on Windows, then a dependency for tzinfo-data will be included. However Rails (up to version 4.1.x) omitted :x64_mingw from the list of platforms and therefore didn't work correctly on 64-bit Windows versions of Ruby.
I found this about tzinfo_data

leecalcote (approved):

Harshit-Dandriyal:
Thanks @leecalcote, I almost forgot about this PR :joy:

PR #724, opened 2022-06-05
Filter count feature

kayceeDev:
**Description**

**Notes for Reviewers**


Gatsby/src/pages/catalog.js
```diff
@@ -8,6 +9,7 @@ import Features from "../components/Catalog/Features/features"
 
 const MembersPage = () => {
   const [filter, setFilter] = useState([]);
+  const [categories] = useState(DataFile)
```
YashKamboj, line 12: @kayceeDev It would be better to use another naming here instead of categories

YashKamboj (changes requested):

Gatsby/src/pages/catalog.js
kayceeDev: What do you suggest?

YashKamboj: "dataFile" can be a good option

YashKamboj: @kayceeDev

kayceeDev: ok. I will send a PR now

YashKamboj (approved):
LGTM!!!

PR #953, opened 2023-01-10
Most Popular discussions posts

kayceeDev:
**Description**

This PR integrates the most popular discussions from the [discuss forum](https://discuss.layer5.io)

**Notes for Reviewers**

- [X] Yes, I signed my commits.


leecalcote:
@asubedy what do you think?

asubedy:
@kayceeDev are we using the same json file for popular post and reset post?

kayceeDev:
> @kayceeDev are we using the same json file for popular post and reset post?

Yes. It has same data. I am just sorted them on number of views.

asubedy:
> > @kayceeDev are we using the same json file for popular post and reset post?
> 
> Yes. It has same data. I am just sorted them on number of views.

I think this should not be the case since the overall popular post could be different from the popular post from the recent topics

kayceeDev:
> > > @kayceeDev are we using the same json file for popular post and reset post?
> > 
> > 
> > Yes. It has same data. I am just sorted them on number of views.
> 
> I think this should not be the case since the overall popular post could be different from the popular post from the recent topics

I think it's same API for posts. Not like different API's for posts.

leecalcote:
So long as the objective is "the most popular posts (by like) with 'meshery' tag", then we're ok. But, if it's not with 'meshery' tag, then, we're misrepresenting.

leecalcote:
Good deal. Ping when you've made your final update. 👍

PR #1043, opened 2023-02-25
mesheryctl installation script updated

deepakdinesh1123:
**Description**
The mesheryctl installation script fetched the latest version of meshery from the releases page by parsing the version from the first line that contains the pattern `/releases/tag` but now the first line does not have a version so the version now fetched from the second line that matches the pattern.
**Notes for Reviewers**


install
```diff
@@ -124,7 +124,7 @@ fi
 
 if [ "x${MESHERY_VERSION}" = "x" ] ; then
   MESHERY_VERSION=$(curl -L -s https://github.com/meshery/meshery/releases/latest | \
-                  grep "/releases/tag/" | head -1 | awk -F '"' '{print $4}' | \
+                  grep "/releases/tag/" | sed -n '2{p;q;}' | awk -F '"' '{print $4}' | \
```
hexxdump, line 127: Lets modify the grep pattern itself to "mesehry/releases/tag"

hexxdump:
```
echo "latest version: "  $(curl -L -s https://github.com/meshery/meshery/releases/latest | \
                  grep "meshery/releases/tag/" | head -1 | awk -F '"' '{print $4}' | \
                  awk -F '/' '{print $NF}' )
```

**`latest version:  v0.6.61`**

Issue #1211, opened 2023-06-22
[Catalog] Missing Descriptions


leecalcote:
Catalog items are missing descriptions

### Contributor Resources

The meshery.io website uses Jekyll and GitHub Pages. Site content is found under the [`master` branch](https://github.com/meshery/meshery.io/tree/master).
- See the [Contributing to Meshery.io Website](https://github.com/meshery/meshery.io#contributing-to-the-mesheryio-website) section of the readme.md and other [contributing instructions](https://docs.meshery.io/project/contributing), too.
- See Meshery site designs in this [Figma designfile](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI?node-id=110%3A1). Fill-in the [Layer5 Community Member Form](https://layer5.io/newcomer) and join the [Layer5 Community Slack](https://slack.meshery.io) for access.

Rajivkumawat3:
I want to work on this issue .  can you assign this issue to me.

Issue #1208, opened 2023-06-22
Error: YAML Exception Reading 


vaibhav-xt:
### Description

On running the command `make site` it shows me an error. A small update needs to both files. 
![Screenshot (86)](https://github.com/meshery/meshery.io/assets/90946899/936b29cc-5ee2-4baa-a13a-f215fbb6bf04)

##### [Here is the link to one of the files:](https://github.com/meshery/meshery.io/blob/master/collections/_catalog/deployment/2df29b2b-7e5e-4d8b-a5b3-3dea569f4d25.md) 
![Screenshot (88)](https://github.com/meshery/meshery.io/assets/90946899/19a79bd3-b5fc-4ffb-80d5-fe2c154b6d53)

### Expected Behavior

Don't show any errors. 

### Enviroment:
 - OS: Mac Linux Windows
 - Browser: Chrome Safari Firefox
 - Version: 
 - Device: Desktop Mobile

---
## Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

The meshery.io website uses Jekyll and GitHub Pages. Site content is found under the [`master` branch](https://github.com/meshery/meshery.io/tree/master).
- 📚 See the [Contributing to Meshery.io Website](https://github.com/meshery/meshery.io#contributing-to-the-mesheryio-website) section of the readme.md and other [contributing instructions](https://docs.meshery.io/project/contributing), too.
- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](https://slack.meshery.io)

vaibhav-xt:
I would like to work on this issue.

Rajivkumawat3:
@vaibhav-xt have you solved YAML error .

vaibhav-xt:
@Rajivkumawat3 yes I solved issue and raised pr.

Rajivkumawat3:
ok because I also had solved this issue

theBeginner86:
We can mark this as completed.
Thanks @vaibhav-xt for raising this issue. 👍

PR #1259, opened 2023-07-04
[CI] Improve Catalog Dispatch Workflow

theBeginner86:
**Description**

This PR fixes https://github.com/meshery/meshery.io/issues/1211

Updates:
1. Ensure workflow runs nightly every 24 hrs.
2. Does cleanup before updating. That is, removes all existing files under [/catalog](https://github.com/meshery/meshery.io/tree/master/catalog) and [ /collection/_catalog](https://github.com/meshery/meshery.io/tree/master/collections/_catalog)
3. Add support multi-line string for pattern caveats and pattern info. Issue: https://github.com/meshery/meshery.io/issues/1208

**Notes for Reviewers**


.github/workflows/add-catalog.yml
```diff
@@ -15,19 +17,29 @@ jobs:
...
+          rm ./collections/_catalog/deployment/*
+          rm ./collections/_catalog/jwt/*
+          rm ./collections/_catalog/observability/*
+          rm ./collections/_catalog/resiliency/*
+          rm ./collections/_catalog/scaling/*
+          rm ./collections/_catalog/security/*
+          rm ./collections/_catalog/traffic-management/*
+          rm ./collections/_catalog/troubleshooting/*
+          rm ./collections/_catalog/wasm/*
+          rm ./collections/_catalog/workloads/*
```
theBeginner86, line 32: @leecalcote 
If this looks good then we can merge and test this out.

The only caveat to deleting all the existing files is that we would also be deleting our old Catalog content which was manually added and is not available in remote provider DB. So we won't necessarily get that. But at least this would make it consistent with https://meshery.layer5.io/catalog

theBeginner86 (commented):
One more thing to note is that currently [categories](https://github.com/meshery/meshery.io/tree/master/catalog/categories), [compatibility](https://github.com/meshery/meshery.io/tree/master/catalog/compatibility), [technology](https://github.com/meshery/meshery.io/tree/master/catalog/technology) is hardcoded and not being tracked based on DB.

So once we start tracking integration spreadsheet data into DB (which we are about to do for connections) then we can have APIs to fetch this specific data and then we can make everything dynamic.

Issue #896, opened 2022-11-03
Website: Consolidate duplicative and unused CSS


leecalcote:
**Current Behavior**
This site has much unused CSS and some overlapping CSS.

**Desired Behavior**
Use of SASS is preferred. Ideally, most of the CSS (by line count) is removed from the site. This can be done piecemeal over multiple PRs.

---
<img src="https://raw.githubusercontent.com/meshery/meshery.io/master/images/logos/meshery-logo.png" width="16px" align="left" /><h4>Contributor [Guides](https://docs.meshery.io/project/contributing) and Resources</h4>

- 🎨 Wireframes and designs for Meshery Site in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🖥 [Contributing to Meshery Website](https://github.com/meshery/meshery.io#contributing-to-the-mesheryio-website)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](http://slack.layer5.io)

tinniaru3005:
Hey @leecalcote, I can work on this. Can you please explain on what do you mean by multiple PRs?

leecalcote:
@tinniaru3005 this effort doesn't have to be completed in one shot.

leecalcote:
@tinniaru3005 does this make sense?

tinniaru3005:
> @tinniaru3005 does this make sense?

Ya, it does. Should I get started with solving it? Is there any specification on breaking up the task that I should follow?

arun-kushwaha04:
Hey @leecalcote can i take this task ?

leecalcote:
Thanks, @arun-kushwaha04

ashutosh887:
@leecalcote Let me try this one now

ashutosh887:
Thanks @leecalcote

DarrenDsouza7273:
@ashutosh887 please update

Ghat0tkach:
I already have some experience in Sass architecture , and upon looking at the repo i found the basic architecture missing as mentioned in 
https://sass-guidelin.es/#architecture , the 7-1  pattern
 Basically, we have all our partials stuffed into 7 different folders, and a single file at the root level (usually named main.scss) which imports them all to be compiled into a CSS stylesheet.

``` base/ ```
```components/```
```layout/```
```pages/```
```themes/```
```abstracts/```
```vendors/```

[...]

PR #1311, opened 2023-07-23
[Enhancement] Updated integrations Section

thisiskaransgit:
**Description**

Changes Done 
- ### Using .map for iteration over traditional for looping for better loading time.
  The use of .map() allows us to transform each element in the data array into its corresponding HTML card, instead of repeatedly updating innerHTML. This results in better performance, as modifying innerHTML multiple times can be heavy.

| Before | After |
|--------|--------|
|   |  |

- ### Remove hardcoding of integration numbers   
  the number displayed currently was hardcoded, even though we had the javascript method declared. 
  
  
### Notes for Reviewers

  TODO: Review the Integration Workflow, we are still at ~210 integrations for the meshery.io website. The updated index number should be ~230.
  
- We need to reduce the overlapping CSS rules( #896 ), which are largely contributing to the load time
- Use Kramdown for blogs and posts 
- Utilizing liquid templating more for reusable components, - Code component, - Features sections.   


vishalvivekm:
@thisiskaransgit Let's discuss this on Websites Meeting call on Monday. Please add this as an agenda item in the [meeting mintes](https://docs.google.com/document/d/1XczAHXVe2FIWPqiF57ospJ43zw5cZQ7ui8mn39v5EvA/edit#heading=h.69qskjv558bk) if you would. :)

leecalcote (approved):

Issue #1337, opened 2023-07-31
Liquid Template for Code Copy Component


thisiskaransgit:
#### Current State of the Meshery.io site

The code component for copying code to clipboard on the site has multiple repetitions of HTML code and is not using the Jekyll Liquid templating system. This can make it difficult to maintain and update the code component across the site.

#### Desired State of the Meshery.io site

Use  Liquid template system for the code copy component on the site. This would allow to reduce the amount of repetitive HTML code and make updates to the component more efficient.

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

The meshery.io website uses Jekyll and GitHub Pages. Site content is found under the [`master` branch](https://github.com/meshery/meshery.io/tree/master).
- 📚 See the [Contributing to Meshery.io Website](https://github.com/meshery/meshery.io#contributing-to-the-mesheryio-website) section of the readme.md and other [contributing instructions](https://docs.meshery.io/project/contributing), too.
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](https://slack.meshery.io)

parth721:
I can work on it

JayKay24:
@parth721 Has already state interest.

leecalcote:
@JayKay24 :(

PR #1348, opened 2023-08-06
Fix liquid template copy component 1337

JayKay24:
**Description**

This PR fixes [#1337 ](https://github.com/meshery/meshery.io/issues/1337)

**Notes for Reviewers**

### DIFF
--- a/_includes/copy-to-clipboard.html
+++ b/_includes/copy-to-clipboard.html
@@ -1,8 +1,8 @@
 <!-- copy to clipboard -->
 <a 
   class="btn tooltip" 
-  data-clipboard-target="{{ clipboard_target }}"
-  data-clipboard-text="{{ clipboard_text }}"
+  data-clipboard-target="{{ include.clipboard_target }}"
+  data-clipboard-text="{{ include.clipboard_text }}"
   onmouseout="resetCopyText(this)" 
   style="position:absolute; top: 0.10rem; right: 0.5rem; font-size: 22px;"
   >
--- a/_includes/copy-to-clipboard.html
+++ b/_includes/copy-to-clipboard.html
@@ -1,0 +1,13 @@
+<!-- copy to clipboard -->
+<a 
+  class="btn tooltip" 
+  data-clipboard-target="{{ clipboard_target }}"
+  data-clipboard-text="{{ clipboard_text }}"
+  onmouseout="resetCopyText(this)" 
+  style="position:absolute; top: 0.10rem; right: 0.5rem; font-size: 22px;"
+  >
+  <i class="far fa-copy"></i>
+  <span class="tooltiptext" style="font-size: 15px; width: 140px; height: 40px; padding: 0; line-height: 40px; top: 2rem; left: -80px;">
+    Copy to clipboard
+  </span>
+</a>
--- a/_includes/help-modal.html
+++ b/_includes/help-modal.html
@@ -111,27 +111,21 @@ Easily <span style="color: #00D3A9; cursor: pointer;" id="myBtn">import any cata
                 <p class="code-box" style="height: 40px; line-height: 15px; color: white"><code style="position: absolute;">
                     mesheryctl pattern apply -f [file | URL]
                  </code></p>
-                  {% assign clipboard_target="#start-pattern" %}
-                  {% assign clipboard_text="mesheryctl pattern apply -f [file | URL]" %}
-                  {% include copy-to-clipboard.html clipboard_target=clipboard_target clipboard_text=clipboard_text %}
+                  {% include copy-to-clipboard.html clipboard_target="#start-pattern" clipboard_text="mesheryctl pattern apply -f [file | URL]" %}
               </div>
             <p>2. Onboard an application.</p>
             <div id="start-app" class="highlight-code">
                 <p class="code-box" style="height: 40px; line-height: 15px; color: white"><code style="position: absolute;">
                     mesheryctl app onboard -f [file-path]
                  </code></p>
-                  {% assign clipboard_target="#start-app" %}
-                  {% assign clipboard_text="mesheryctl app onboard -f [file-path]" %}
-                  {% include copy-to-clipboard.html clipboard_target=clipboard_target clipboard_text=clipboard_text %} 
+                  {% include copy-to-clipboard.html clipboard_target="#start-app" clipboard_text="mesheryctl app onboard -f [file-path]" %} 
               </div>
             <p>3. Apply a WASM filter file.</p>
             <div id="start-pattern" class="highlight-code">
                 <p class="code-box" style="height: 40px; line-height: 15px; color: white"><code style="position: absolute;">
                     mesheryctl exp filter apply --file [GitHub Link]
                  </code></p>
-                  {% assign clipboard_target="#start-pattern" %}
-                  {% assign clipboard_text="mesheryctl exp filter apply --file [GitHub Link]" %}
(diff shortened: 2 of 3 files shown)


Ghat0tkach:
hey @JayKay24 ,

Let's discuss it on the websites call.
Please add this as an agenda item in the meeting minutes, if you would :)

https://docs.google.com/document/d/1XczAHXVe2FIWPqiF57ospJ43zw5cZQ7ui8mn39v5EvA/edit#heading=h.lohhtewfwima

thisiskaransgit (commented):
@JayKay24, thanks much, please take a look into the `footer.html` to make sure there's no overlapping script.

JayKay24:
@thisiskaransgit Thanks for your review. The code in [footer.html](https://github.com/meshery/meshery.io/blob/a2002a46a2b3ab434f3c439a7aa26384b43639fc/_includes/footer.html#L75C1-L75C1) only has the javascript function(`resetCopyText`) and dependency(`clipboard.min.js`) needed to copy a piece of text. It is only defined in that file. I abstracted away the repeated HTML into a `copy-to-clipboard.html` template that still references the `resetCopyText` defined in `footer.html`.

thisiskaransgit (approved):
We are good to then, thanks @JayKay24 👍🏼

JayKay24:
I've updated the branch to keep it mergable but will need a re-review @thisiskaransgit

thisiskaransgit (commented):
@JayKay24, use of CDN, we need to replace it with a locally available file, otherwise the payload will not be improved. You can reference the implementation in docs.meshery.io if required. I overlooked it initially but we need this update

JayKay24:
@thisiskaransgit does that mean that I integrate `npm` into the project? This would allow me to use the clipboard library currently loaded via CDN as a local npm module.

thisiskaransgit:
> @thisiskaransgit does that mean that I integrate `npm` into the project? This would allow me to use the clipboard library currently loaded via CDN as a local npm module.

@JayKay24 declare the javascript logic in the code. the desired behavior should be that the class element can use the locally hosted code and not third-party module, which will a) reduce load b) filter the unused code that comes along c) opt out of depending on other services. This is what I intend to update for other sections too actually. 

take a look at here for reference:  https://github.com/meshery/meshery/blob/master/docs/assets/js/main.js

Chadha93:
> > @thisiskaransgit does that mean that I integrate `npm` into the project? This would allow me to use the clipboard library currently loaded via CDN as a local npm module.
> 
> @JayKay24 declare the javascript logic in the code. the desired behavior should be that the class element can use the locally hosted code and not third-party module, which will a) reduce load b) filter the unused code that comes along c) opt out of depending on other services. This is what I intend to update for other sections too actually.
> 
> take a look at here for reference: [meshery/meshery@`master`/docs/assets/js/main.js](https://github.com/meshery/meshery/blob/master/docs/assets/js/main.js?rgh-link-date=2023-08-20T18%3A20%3A34Z)

@JayKay24, does this helps?

JayKay24:
@Chadha93 i haven't had a chance to dive deep into the solution but from what I can tell is that I will wrap the clipboard logic inside an IIFE. Will look to address this tomorrow. I'll post any questions I have in this thread.

JayKay24:
@thisiskaransgit The example you gave me also references the `clipboard.min.js` library via CDN in this file -> https://github.com/meshery/meshery/blob/master/docs/_layouts/default.html.

I don't understand what you mean when you say the payload will be improved. Which payload is this? Do you mean that what we want is to avoid the extra HTTP request of downloading the `clipboard.min.js` dependency on page load in order to achieve a faster loading rate by having the browser parse a  local `.js` file with clipboard logic?
 If this is the case, there's a way to use the dependency locally by downloading and [extracting the zip file](https://www.npmjs.com/package/clipboard) if we're not into package management. But if we do it this way, we won't have a way to keep it up to date easily.
The other option is to integrate the project with `npm` and reference `clipboard.js` as a node module.  This way when the project is built, the dependency will be included on page load without any extra HTTP request to the CDN.
Please help me to clarify this matter. Sorry If I'm not understanding it as quickly.
@Chadha93 please let me know your insights on this. What could I be missing?

_includes/help-modal.html
```diff
@@ -111,36 +111,21 @@ <h3>Meshery CLI</h3>
                 <p class="code-box" style="height: 40px; line-height: 15px; color: white"><code style="position: absolute;">
                     mesheryctl pattern apply -f [file | URL]
                  </code></p>
-                  <!-- copy to clipboard -->
-                  <a class="btn tooltip" style="position: absolute; top:0.10rem; right: 0.5rem; font-size: 22px;" data-clipboard-target="#start-pattern"
-                    data-clipboard-text="mesheryctl pattern apply -f [file | URL]" onmouseout="resetCopyText(this)">
-                    <i class="far fa-copy"></i>
-                    <span class="tooltiptext" style="font-size: 15px; width: 140px; height: 40px; padding: 0; line-height: 40px; top: 2rem; left: -80px;">Copy to clipboard</span>
-                  </a>
+                  {% include copy-to-clipboard.html clipboard_target="#start-pattern" clipboard_text="mesheryctl pattern apply -f [file | URL]" %}
```
vishalvivekm, line 114:
```suggestion
<!-- copy to clipboard -->
                  {% include copy-to-clipboard.html clipboard_target="#start-pattern" clipboard_text="mesheryctl pattern apply -f [file | URL]" %}
```

_includes/help-modal.html
```diff
@@ -111,36 +111,21 @@ <h3>Meshery CLI</h3>
...
                 <p class="code-box" style="height: 40px; line-height: 15px; color: white"><code style="position: absolute;">
                     mesheryctl exp filter apply --file [GitHub Link]
                  </code></p>
-                  <!-- copy to clipboard -->
-                  <a class="btn tooltip" style="position: absolute; top:0.10rem; right: 0.5rem; font-size: 22px;" data-clipboard-target="#start-pattern"
-                    data-clipboard-text="mesheryctl exp filter apply --file [GitHub Link]" onmouseout="resetCopyText(this)">
-                    <i class="far fa-copy"></i>
-                    <span class="tooltiptext" style="font-size: 15px; width: 140px; height: 40px; padding: 0; line-height: 40px; top: 2rem; left: -80px;">Copy to clipboard</span>
-                  </a>
+                  {% include copy-to-clipboard.html clipboard_target="#start-pattern" clipboard_text="mesheryctl exp filter apply --file [GitHub Link]" %}
```
vishalvivekm, line 128:
```suggestion
<!-- copy to clipboard -->
                  {% include copy-to-clipboard.html clipboard_target="#start-pattern" clipboard_text="mesheryctl exp filter apply --file [GitHub Link]" %} 
```

_includes/copy-to-clipboard.html
```diff
@@ -0,0 +1,13 @@
+<!-- copy to clipboard -->
+<a 
+  class="btn tooltip" 
+  data-clipboard-target="{{ clipboard_target }}"
```
vishalvivekm, line 4:
```suggestion
  data-clipboard-target="{{ include.clipboard_target }}"
```

_includes/copy-to-clipboard.html
```diff
@@ -0,0 +1,13 @@
...
+  data-clipboard-text="{{ clipboard_text }}"
```
vishalvivekm, line 5:
```suggestion
  data-clipboard-text="{{ include.clipboard_text }}"
```

vishalvivekm (approved):
Looks good to me.
Thanks @JayKay24 🚀

PR #1761, opened 2024-05-28
replace - with space

iArchitSharma:


_includes/patterns-filter.html
```diff
@@ -331,7 +331,7 @@
     let whiteImg = data[i].white;
 
     label.innerHTML = `<img src="${colorImg}" id="logo-dark-light" data-logo-for-dark="${whiteImg ?? colorImg}"
-                    data-logo-for-light="${colorImg}" /><div class="name">${data[i].name}</div><div class="badge">${count}</div>`;
+                    data-logo-for-light="${colorImg}" /><div class="name">${data[i].name.replace(/-/g, " ")}</div><div class="badge">${count}</div>`;
```
Yashsharma1911, line 334: @iArchitSharma We don't need string manipulation, is name only property we get in data? do we get displayName? I will say we have can get displayName please use the instead doing string manipulation with name

iArchitSharma: @Yashsharma1911 we don't have any displayName or similar property
image

PR #1784, opened 2024-06-27
[Catalog] Add artifacthub-pkg version and displayname

dusdjhyeon:
**Description**

This PR fixes #
If that pr https://github.com/meshery/meshkit/pull/518/files
works properly, artifacthub-pkg.yml would be able to have both version and displayname. 
**Notes for Reviewers**


assets/artifact-hub-pkg/package.go
```diff
@@ -190,7 +192,22 @@ func writePatternFile(pattern CatalogPattern, patternType, patternInfo, patternC
...
+		version = semver.New(0, 0, 1, "", "").String()
+	} else {
+		v, err := semver.NewVersion(version)
+		if err != nil {
+			return meshkitErrors.New(ErrInvalidVersionCode, meshkitErrors.Alert,
+				[]string{"Failed to parse the version"},
+				[]string{fmt.Sprintf("The provided version '%s' is invalid.\nError: %v", version, err)},
+				[]string{"The version string is not in a valid semantic version format"},
+				[]string{"Ensure the version string follows semantic versioning format", "Check for typos or incorrect format"})
+		}
```
MUzairS15, line 206: No need to increment the version.
As `published_version` is the version user has selected to publish.
Increment will lead to ambiguity/discrepancy.

Also, we are sure that the `published_version` will be a valid semver version, otherwise the designs do not get persist. So we can remove the check here.

MUzairS15 (changes requested):

MUzairS15:
New Meshkit is released, you can bump the meshkit version.

assets/artifact-hub-pkg/package.go
leecalcote: @MUzairS15, in order for this to be true our existing data must be cleaned up. Right, @Yashsharma1911?

dusdjhyeon: I see, I understand. Should I delete the code and test it after existing data is cleaned up?

dusdjhyeon:
This is the new result https://github.com/dusdjhyeon/meshery.io/blob/for-test-3/catalog/b2b05f54-fd48-44f5-bf2d-308761874d59/artifacthub-pkg.yml

Now, we have version and displayName. Is there anything missing?

MUzairS15 (approved):

PR #1796, opened 2024-07-03
Decode pattern info and caveats before adding to ArtifactHub package file

shubham251972:
**Description**

This PR fixes #

Catalog content is in URI format and during artifact-hub package generation we add content without decoding it, which causes issue of non readable content getting published on artifact hub, this PR fixes this issue by making sure we first decode it and then add it to artifact hub package file

yaml.marshal will take care of multi line and special characters

**Notes for Reviewers**


leecalcote (approved):
LGTM

PR #1813, opened 2024-07-08
 Enhance Artifact Hub Package Generation to Support Multiple Versions

Yashsharma1911:
### Description
This PR introduces enhancements to the Artifact package generation process, enabling support for generating multiple versions dir, below is example of dir structure change

This PR also addresses the issue of setting the createdAt property new with the current time on every run of add catalog workflow, which was causing excessive and unnecessary differences in commit. Instead, createdAt is now sourced from the API data.

```
catalog
├── artifacthub-repo.yml
├── designId
│   ├── 1.0.0
│   │   ├── artifacthub-pkg.yml
│   │   └── design.yml
│   └── 2.0.0
│       ├── artifacthub-pkg.yml
│       └── design.yml
└── designId
    └── 1.0.0
        ├── artifacthub-pkg.yml
        └── design.yml
```
**Notes for Reviewers**

Number of changes is high because of dir structure change


assets/artifact-hub-pkg/package.go
```diff
@@ -122,7 +123,22 @@ func processPattern(pattern CatalogPattern, token string) error {
...
+		version = semver.New(0, 0, 1, "", "").String()
+	}
+
+	// Ensure the version directory exists within the catalog file path, creating it if necessary
+	catalogFilePath := filepath.Join("..","..", mesheryCatalogFilesDir, pattern.ID)
+	versionDir := filepath.Join(catalogFilePath, version)
+    if _, err := os.Stat(versionDir); os.IsNotExist(err) {
+		err = os.MkdirAll(versionDir, 0755)
+		if err != nil {
+			return err
```
dusdjhyeon, line 137: IMO, It seems better to return meshkit err. The `writePatternFile` below already returns meshkit err, which is why it returns err directly.

assets/artifact-hub-pkg/package.go
```diff
@@ -198,8 +212,14 @@ func writePatternFile(pattern CatalogPattern, patternType, patternInfo, patternC
 		patternImageURL = "/assets/images/logos/service-mesh-pattern.svg"
 	}
 
-	format := "2006-01-02 15:04:05Z"
-	currentDateTime, err := time.Parse(format, time.Now().UTC().Format(format))
+	parsedTime, err := time.Parse(time.RFC3339Nano, pattern.CreatedAt)
+	if err != nil {
+		return err
```
dusdjhyeon, line 217: If this is also possible, I would suggest using meshkit err.

assets/artifact-hub-pkg/package.go
leecalcote: Good catch!

Yashsharma1911:
Thanks for reviewing I'm merging this PR

PR #1833, opened 2024-07-29
[CI] Pull Changes Before `auto-commit` Action

humblenginr:
**Notes for Reviewers**

This PR fixes pulls the changes from the repository before running the [`auto-commit`](https://github.com/stefanzweifel/git-auto-commit-action) action. This is to make sure that there are no new changes introduced in the time between when the code was checked out and when the workflow is trying to commit the changes. 

This PR ensures that the code is being pulled before running the `auto-commit` action in all of Meshery.io's workflows. 

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**


leecalcote (approved):
This is important!

PR #2019, opened 2025-01-13
Error Code Update workflow

souvikinator:
**Description**

This PR fixes #

This should add the error code updater workflow for the `artifact-hub-pkg` and running the workflow would update the docs as well

**Notes for Reviewers**


leecalcote (changes requested):
There is no such word as "updation". Please correct.

PR #2072, opened 2025-03-09
[CI] Update runner image and fix error in workflow

shinigami-777:
**Description**

The PR #2068 had a workflow error for which `ruby/setup-ruby@v1.162.0` was the cause.
Changing it to `ruby/setup-ruby@v1` seems to solve that problem.

@sangramrath I've tested the updates using a different branch.

### DIFF
--- a/.github/workflows/add-catalog.yml
+++ b/.github/workflows/add-catalog.yml
@@ -7,7 +7,7 @@ on:
     types: [add_catalog]
 jobs:
   UpdateCloudNativeCatalog:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     outputs:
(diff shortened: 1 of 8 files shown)


leecalcote:
This is a bit odd. What's the actual reason for the failure? We will have to move forward at some point. Please open an issue to track this, @shinigami-777.

shinigami-777:
> This is a bit odd. What's the actual reason for the failure? We will have to move forward at some point. Please open an issue to track this, @shinigami-777.

Created an issue #2074 .

sangramrath:
@shinigami-777 can you check with a newer release of `ruby/setup-ruby` like the latest one being `1.227.0` or a version more than 1.162.0 which is failing.

shinigami-777:
Changing it to the latest `1.227.0` version works. I've checked it using a different workflow.

Issue #2444, opened 2025-12-05
Anchor tag not scrolling to community forum section


marblom007:
### Description
Clicking this link, https://meshery.io/community#discussion-forums/, does not result in browsers scrolling to the respective section on the page.

---
## Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community/handbook)

The meshery.io website uses Jekyll and GitHub Pages. Site content is found under the [`master` branch](https://github.com/meshery/meshery.io/tree/master).
- 📚 See the [Contributing to Meshery.io Website](https://github.com/meshery/meshery.io#contributing-to-the-mesheryio-website) section of the readme.md and other [contributing instructions](https://docs.meshery.io/project/contributing), too.
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](/community#discussion-forums) and [Community Slack](https://slack.meshery.io)

denishjackson1:
@marblom007, kindly assign this to me. I have already worked on it, and I would like to raise the PR.

Thank you.

GaneshPatil7517:
@marblom007 Please assign this task to me. I’ve already completed the work and would like to raise a PR.

kishore08-07:
@denishjackson1 @GaneshPatil7517 
then one of you can go ahead and open a PR, and the other can review it.

SujalTripathi:
@marblom007 is these issue open so please assign these to me!

Ishaparte:
@marblom007 can you pls assign this issue to me

KhushamBansal:
hi @saurabhraghuvanshii , can you assign me this issue?

saurabhraghuvanshii:
Thanks @KhushamBansal but issue has been resolved

PR #2445, opened 2025-12-06
Fix anchor navigation for the Discussion Forums section on the community page

GaneshPatil7517:
This PR fixes the scrolling issue described in meshery/meshery.io#2444.
Navigating to /community#discussion-forums did not scroll the page to the correct section because the heading landed behind the fixed header. This update corrects the anchor behavior so the section scrolls into clear view.

Notes for Reviewers

Updated the discussion-forums section to ensure the anchor scrolls properly.
Added an appropriate scroll offset / ensured correct id so the section is fully visible when navigated via anchor link.
o layout or structural changes were made beyond fixing this navigation issue.
Tested across standard desktop and mobile views.

How to Test

Open the following URL in a browser:
https://meshery.io/community#discussion-forums

The page should scroll smoothly.
The “Discussion Forums” heading should appear clearly below the header, without being hidden.

[Signed commits](https://docs.meshery.io/project/contributing#signing-off-on-commits-developer-certificate-of-origin)
 Yes, I signed my commits.

PR fixes # 2444


Fireentity (commented):
LGTM!

denishjackson1:
Based on issue #2444, this PR does not fully resolve the problem. When navigating to https://meshery.io/community#discussion-forums/ the page does not scroll to the discussion-forums section as expected.

Fireentity:
@denishjackson1 You're right. I tested this without the "/" at the end. Why do we need it to work with the "/" at the end, though?

denishjackson1:
> @denishjackson1 You're right. I tested this without the "/" at the end. Why do we need it to work with the "/" at the end, though?

You’re right that the trailing “/” isn’t needed after the "#". After removing it, the current live page scrolls smoothly to the discussion-forums section. However, if the trailing “/” must be supported, a simple script can handle that case.

PR #2660, opened 2026-05-03
Fix Liquid append syntax warnings

arjunmehta-git:
## Summary
- replace the invalid parenthesized Liquid filter in `blog/index.html` with an intermediate slug assignment
- apply the same Liquid-safe pattern in `_includes/partials/individual-component.html`
- remove the Jekyll build warning caused by `append: (value | slugify)` expressions

## Testing
- bundle exec jekyll build


arjunmehta-git:
@copilot implement feedback in the comments on this PR.

PR #2731, opened 2026-06-15
academy do

sarajkrishnasingh:
- **Makefile: prefer Docker for make site; avoid automatic bundle install when Docker not running; update README note**
- **Extensions: add digitalocean academy**


ritzorama (approved):
great!

PR #2744, opened 2026-06-18
fix: add missing meshery-arrows model and persist catalog filter state

### LINKED ISSUE #2743: [Catalog] Missing meshery-arrow model and non-persistent filter on page change
### Description
I had mentioned two issue of frontend which can be updated to make the frontend better.
1) Once we select any filter it shows the required result but as we open the design or models and back it, it reset to its original one .
2) The Meshry arrow link is broken in the oracle design is broken .

### Expected Behavior
1)If I am applying any filter then it should not be changed until I removed it .
2) the model of meshery arrow should open on clicking that.

I am attaching a video recording showing both the issues

https://github.com/user-attachments/assets/8e12feb6-4764-4294-8c62-3bff9e86797e

### Environment:
 - OS: Mac Linux Windows

---

- 🙋🏾🙋🏼 Questions: [Discussion Forum](/community#discussion-forums) and [Community Slack](https://slack.meshery.io)

dhruveshmishra:
This PR fixes #2743 

I have resolved two key issues in the catalog pages of `meshery.io`:

1. **Fixed the Broken `meshery-arrows` Link & Images**: I added the missing model collection file of mesheryarrow . This fixes the broken link when clicking the "meshery-arrows" model button.

2. **Implemented Filter State Persistence**: I modified the filter logic to store the user's checked categories and technologies in `sessionStorage`. Now, when navigating back to the catalog page, the filter checkboxes and lists automatically sync and display the previously selected items instead of resetting to the default state.
Contributing Conventions:

I am attaching the Screen recording of my local server showing the fixex I have made 

https://github.com/user-attachments/assets/2984f44b-648e-450a-b6fc-9e1120588d28

https://github.com/user-attachments/assets/7f5bf6cf-b35c-47b6-a8df-92e19727f5d8

## Summary by CodeRabbit

* **New Features**
  * Added the Meshery Arrows model with descriptive information and visual flow annotation guidance.
  * Added dedicated long-arrow icon support for Meshery Arrows, with a fallback image when needed.
  * Filter selections for technologies, categories, and pattern types now persist per page during the same browser session and are automatically restored after refreshing or revisiting the page.


_includes/patterns-filter.html
```diff
@@ -227,19 +227,63 @@
...
 <script type="module">
   document.addEventListener("DOMContentLoaded", () => {
-  // Uncheck all checkboxes on page load AND reset their state in the DOM
-  const checkboxes = document.querySelectorAll("input[type='checkbox']");
-  checkboxes.forEach((checkbox) => {
-    checkbox.checked = false;
-  });
+    // Wait a brief moment for browser state restore/autocomplete to complete
+    setTimeout(() => {
+      // 1. Sync from DOM in case browser restored checkmarks
```
banana-three-join: Step one has a lot of repetition. The behavior can be abstracted away in a helper function instead.

_includes/patterns-filter.html
```diff
@@ -227,19 +227,63 @@
...
+          }
+        });
+        sessionStorage.setItem('selected_patterntypes', JSON.stringify(globalPatternTypeList));
+      }
 
-  // Reset global filter lists
-  globalTechnologyList = [];
-  globalCategoryList = [];
-  globalPatternTypeList = [];
+      // 2. Make sure DOM checkboxes are checked for whatever is in the lists (loaded from sessionStorage)
```
banana-three-join: Same with step two. It can be abstracted away in a helper function to avoid repetition.

_includes/patterns-filter.html
```diff
@@ -227,19 +227,63 @@
...
+      categoryInputs.forEach(input => {
+        if (globalCategoryList.includes(input.value)) {
+          input.checked = true;
+        } else {
+          input.checked = false;
+        }
+      });
+      techInputs.forEach(input => {
```
banana-three-join: Technologies can also be undefined because it's not present in all of the pages.

banana-three-join (changes requested):
Aside from my feedback, please reply to each individual feedback given by the LLM and not just mark it as resolved.

dhruveshmishra:
@banana-three-join  sorry for that I had checked the llm feedback and after working on that I had pushed it into my branch now you can take a look

Bhumikagarggg (dismissed):
LGTM 🚀

dhruveshmishra:
@Bhumikagarggg  Thanks mam

dhruveshmishra:
@Bhumikagarggg  @banana-three-join  any update for this pr

_includes/patterns-filter.html
dhruveshmishra: I had assumed the technology attribute would always be defined on all pages. 
i corrected this by adding a safety check before parsing: `const technologies = technologyAttr ? JSON.parse(technologyAttr) : [];`. This ensures that pages without technologies are handled gracefully and do not throw undefined errors.

_includes/patterns-filter.html
dhruveshmishra: Originally, this manual unchecking/checking block was used to reset states. i have removed it entirely since states are now initialized cleanly from session storage when checkboxes are generated.

_includes/patterns-filter.html
dhruveshmishra: Originally, the timeout was added to wait for the browser's autocomplete on load. i have now removed the timeout and manual sync loop entirely, setting the checkbox states directly during their dynamic creation.

Bhumikagarggg:
@dhruveshmishra Thank you for your contribution! Let's discuss this during the website call tomorrow at 5:30 PM IST | 7 AM CST Add it as an agenda item to the [meeting minutes](https://meet.layer5.io/websites-minutes), if you would :slightly_smiling_face:

dhruveshmishra:
@Bhumikagarggg ohk mam

images/integrations/meshery-arrows/icons/color/meshery-arrows-color.svg
saurabhraghuvanshii, line 1: @dhruveshmishra  did you add this svg manually?

saurabhraghuvanshii (dismissed):

images/integrations/meshery-arrows/icons/color/meshery-arrows-color.svg
dhruveshmishra: Yes, I added these svgs manually. Since the meshery arrows model was missing and is being introduced in this PR and as per my knowledge I cannot find any existing icons in the repository.

_includes/patterns-filter.html
```diff
@@ -446,13 +442,16 @@
...
     }
+    sessionStorage.setItem('selected_technologies_' + pageKey, JSON.stringify(globalTechnologyList));
   }
 
   // This function updates the category list
   function updateCategoryList(value) {
     // When user checks a category, that category category is pushed in globalCategoryList array which holds the categories selected
     if (document.getElementById(value).checked) {
-      globalCategoryList.push(value);
+      if (!globalCategoryList.includes(value)) {
```
banana-three-join: I don't think the  if (!globalCategoryList.includes(value)) { } safeguard is needed. If there's any issues with how the global array is being handled then the logic should be fixed within the handling of the array.

banana-three-join (changes requested):
You're not updating the initial load of the pattern list to see whether a value exists in the array or not to mark its element's input as checked

_includes/patterns-filter.html
dhruveshmishra: The check !globalCategoryList.includes(value) was originally added as a defensive safeguard to ensure we didn't end up with duplicate entries in the array.

However, since the push logic is strictly tied to the checkbox's .checked state, it's impossible to trigger two consecutive checks without an uncheck in between. I have removed the redundant safeguard to keep the array handling clean. Thanks for pointing it out.

banana-three-join:
> You're not updating the initial load of the pattern list to see whether a value exists in the array or not to mark its element's input as checked

This is still not being addressed.  @dhruveshmishra

dhruveshmishra:
@banana-three-join My apologies! I only saw your second review comment about the safeguard check and missed this.
Now I had updated the pattern list loop

dhruveshmishra:
@saurabhraghuvanshii I had updated, by deleting the `meshery-arrows` svg file to file existed in newcomers `long-arrow.svg` asset.

_includes/patterns-filter.html
```diff
@@ -349,6 +339,9 @@
     //     updatePatternTypeList(cardPatternType[i]);
     //     showFilterCards();
     //   };
+    //   if (globalPatternTypeList.includes(cardPatternType[i])) {
```
banana-three-join, line 417: Why was this commented out?

banana-three-join (changes requested):

_includes/patterns-filter.html
dhruveshmishra: Since the Pattern Type filter is currently disabled/commented out in the UI, I commented out this logic as well. I kept the fully working code here so it can easily be uncommented if we decide to enable the Pattern Type filter in the future

banana-three-join: But above the structure for it is still kept for the pattern list over in line 276. Also, in line 475, you're setting the storage for it.

dhruveshmishra: I kept the variable structure and sessionstorage logic active so the filter would be fully ready to go if we decide to uncomment the UI in the future.

However, I see your point because the checkboxes are commented out, the script still reads from sessionstorage and filters the cards on page load, meaning the user could get stuck with filtered cards and no way to reset them.

dhruveshmishra:
@banana-three-join @Bhumikagarggg @rishiraj38 please review this pr, if any further changes is needed in this

_includes/partials/compatibility.html
```diff
@@ -48,6 +48,8 @@
     <img class="service-mesh-image" src="{{site.baseurl}}/assets/images/integration/keda-http-scaler/icons/color/keda-http-scaler-color.svg" />
     {% elsif item == "wordpress-operator" %}
     <img class="service-mesh-image" src="{{site.baseurl}}/assets/images/integration/wordpress-operator/icons/color/wordpress-operator-color.svg" />
+    {% elsif item == "meshery-arrows" %}
+    <img class="service-mesh-image" src="{{site.baseurl}}/assets/images/newcomers/long-arrow.svg" />
```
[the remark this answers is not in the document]
dhruveshmishra: fixed

banana-three-join:
Please squash your commits. 25 commits for this fix is too excesive. Once that's done, I'll review it once again and fix any standing errors.

dhruveshmishra:
@banana-three-join 
Done! Squashed all commits into a single clean commit.

banana-three-join (changes requested):
Okay. You need to be more thoughtful with what you're pushing in your PRs. 

The scope of the PR and the issue itself is, the missing meshery-arrows and the lack of persistence in the catalog filter. So, why is there changes to the calendar, discuss, charts and such? Please fix that.

This PR should only include the addition of the `meshery-arrows` and the persistence of the filter. On another note, where did you get the definition for the `meshery-arrows` model?

banana-three-join:
Please, look into the model definition over at https://github.com/meshery/meshery/tree/master/models/meshery-arrows/0.7.2/v1.0.0 as well

Let me know if you have any questions.

dhruveshmishra:
@banana-three-join Sorry lenox I was working on some other issue and accidently pushed these files, I had noticed now that it has been pushed along my after commits, I will revert that

dhruveshmishra:
@banana-three-join  ^^

banana-three-join (changes requested):
A few things to work through before this can merge. I've grouped them by subject so you can tackle one at a time.
 
---
 
**1. There's already a source of truth for this model in the repo**
 
Before touching the metadata: `assets/modelsFiles/meshery-arrows.tar` is already committed here. Unpack it and you get:
 
```
0.7.2/v1.0.0/model.json
0.7.2/v1.0.0/components/*.json   (25 files)
```
 
`model.json` gives you the canonical values for this model, and `metadata.svgColor` / `metadata.svgWhite` contain the real icons as inline SVG — two distinct ones, teal `#00B39F` and white. Each of the 25 component JSONs likewise carries its own `styles.svgColor` and `styles.svgWhite`.
 
The `.md` still has to be written by hand — nothing here generates it — but every value in it should be **copied from that tar** rather than authored from scratch. That also answers @saurabhraghuvanshii's question about where the SVGs came from.
 
One thing that's already working, for what it's worth: the "Download Model" button in `_layouts/single-page-model.html:172` resolves `/assets/modelsFiles/<name>.tar`, and that file exists — so the download on your new page works with no extra effort.
 
---
 
**2. Model metadata — `collections/_models/meshery-arrows/meshery-arrows.md`**
 
The file to copy the shape from is `collections/_models/meshery-core/meshery-core.md`. It's the closest sibling: same registrant (`meshery`), same category, same annotation-model family, already merged.
 
| field | current | should be |
|---|---|---|
| `subcategory` | `Uncategorized` | `Application Definition & Image Build` |
| `componentsCount` | `0` | `25` |
| `components` | `[]` | list of the 25 components |
| `relationships` | `[]` | blank |
| `docURL` | `/concepts/logical/models` | `/extensibility/integrations/meshery-arrows` |
| `subtitle` | custom | `Collaborative and visual infrastructure as design for Meshery Arrows` |
 
Two of these are functional, not cosmetic:
 
**`components: []` will render a broken section.** Empty arrays are *truthy* in Liquid, so the `{% if page.components %}` guard at `_layouts/single-page-model.html:209` passes and the page emits a `<h3>Components</h3>` heading with nothing beneath it. None of the ~400 files in `_models` use `[]` — `collections/_models/aigisuk/aigisuk.md` and `collections/_custom-models/_custom-model-definition-template.md` both leave the key blank. Same applies to `relationships`.
 
**`componentsCount: 0` is wrong** — the model ships 25 components (`components_count: 25` in `model.json`, and 25 files in `components/`).
 
On listing all 25: I checked, and there is no precedent in this repo for `componentsCount > 0` with a blank `components` list — every model with components enumerates them. So the list does need writing out, following the meshery-core format. It's tedious, but the names and icons all come straight out of the tar rather than being invented.
 
One field that's already right: `description: Icons that used for pointing out` matches the registry exactly. CodeRabbit's suggestion to add "are" would make this file diverge from `meshery/meshery` — if the grammar bothers us, it should be fixed upstream first and mirrored here.
 
---
 
**3. Icons**
 
Right now both `colorIcon` and `whiteIcon` point at `/assets/images/newcomers/long-arrow.svg`, and `compatibility.html` points at the same file.
 
Two problems with that. It couples a model icon to a decorative asset that belongs to the newcomers page, and it uses one green arrow for both themes — `_layouts/single-page-model.html:119` feeds `whiteIcon` into `data-snapshot-dark`, and the registry ships a real white variant we're not using.
 
`meshery-core` has its icons in **two** trees, which is exactly why it needs no special-casing anywhere:
 
```
assets/images/integration/meshery-core/icons/color/meshery-core-color.svg
assets/images/integration/meshery-core/icons/white/meshery-core-white.svg
images/integrations/meshery-core/icons/color/meshery-core-color.svg
```
 
Writing `metadata.svgColor` and `metadata.svgWhite` from the tar into the equivalent `meshery-arrows` paths covers both consumers — the model page front matter and the compatibility fallback.
 
If you do enumerate the components, their icons follow `assets/images/integration/meshery-arrows/components/<component>/icons/{color,white}/`, and all 25 pairs are in the component JSONs.
 
---
 
**4. `_includes/partials/compatibility.html`**
 
Once the icon from section 3 exists at `images/integrations/meshery-arrows/icons/color/meshery-arrows-color.svg`, the generic `{% else %}` branch resolves it and **the new `{% elsif %}` isn't needed at all**.
 
The file's own comment says *"Please remove these hardcoding"*, and this PR adds one more entry to that list. Dropping the branch and adding the icon file is close to what you had originally — the issue was the hand-drawn SVG, not the location you put it in.
 
---
 
**5. `_includes/patterns-filter.html`**
 
The persistence logic is good, and setting `input.checked` at creation time is the right resolution to the earlier `setTimeout` feedback. Two things left:
 
**The `JSON.parse` calls sit at module top level and aren't guarded.** A malformed stored value, or a `SecurityError` when `sessionStorage` is unavailable, throws during module evaluation. Since the checkbox construction lives in that same module, the failure mode isn't "filters don't restore" — it's **no filter checkboxes render at all**. Wrapping it also gives you the abstraction I asked for twice earlier:
 
```js
const pageKey = window.location.pathname;
 
function loadFilter(name) {
  try { return JSON.parse(sessionStorage.getItem(`selected_${name}_${pageKey}`)) || []; }
  catch { return []; }
}
function saveFilter(name, list) {
  try { sessionStorage.setItem(`selected_${name}_${pageKey}`, JSON.stringify(list)); }
  catch { /* storage unavailable */ }
}
 
let globalTechnologyList = loadFilter('technologies');
let globalCategoryList   = loadFilter('categories');
```
 
**The pattern-type persistence is dead code.** The pattern-type checkbox block is entirely commented out, so `updatePatternTypeList` never runs and `selected_patterntypes_*` is never written. The lines you added inside that comment block are noise. Either drop the pattern-type handling from this PR, or reinstate it together with the block — but don't leave half of it live and half commented.
 
Optional, not blocking: module scripts execute before `DOMContentLoaded`, so the listener only costs an extra tick during which unfiltered cards are briefly visible. Calling `showFilterCards()` at the end of the module body removes the flicker Gemini flagged.
 
---
 
**6. Scope**
 
`meshery-shapes` sits in the same design's `compatibility` list with an identical problem — tar present in `assets/modelsFiles/`, no page in `_models`, no icon. Same story for `meshery-flowchart`, `meshery-dev-icons`, and `meshery-simple-line-icons`.
 
Please **don't** pull those into this PR. Given the earlier out-of-scope files, keep this one to `meshery-arrows` plus the filter persistence, and open a follow-up issue for the rest if you'd like to pick them up next.

dhruveshmishra:
@banana-three-join Thanks for the thorough review! I've addressed all points and also attached the screen recording :


1. **Source of Truth & SVGs**: Extracted canonical model icons (`svgColor` and `svgWhite`) and all 25 component SVGs directly from `assets/modelsFiles/meshery-arrows.tar`.
2. **Model Metadata**: Updated `collections/_models/meshery-arrows/meshery-arrows.md` with:
   - `subcategory: Application Definition & Image Build`
   - `componentsCount: 25` and enumerated all 25 components
   - Removed `[]` for `relationships` (left key blank)
   - `docURL: https://docs.meshery.io/extensibility/integrations/meshery-arrows`
   - `subtitle: Collaborative and visual infrastructure as design for Meshery Arrows`
3. **Icons & Fallback**: Added official color/white SVGs to `assets/images/integration/meshery-arrows/` and `images/integrations/meshery-arrows/`.
4. **Compatibility Partial**: Removed the hardcoded `meshery-arrows` branch from `_includes/partials/compatibility.html`.
5. **Filter Persistence**: Wrapped `sessionStorage` in guarded `loadFilter` / `saveFilter` helper functions with `try...catch`, cleaned up dead pattern-type code, and invoked `showFilterCards()` at the module end to prevent card flicker.
6. **Scope**: Kept strictly scoped to `meshery-arrows` and filter persistence.

Please have a look on that

collections/_models/meshery-arrows/meshery-arrows.md
```diff
@@ -0,0 +1,124 @@
...
+  description: 
+componentsCount: 25
+relationships: 
+relationshipsCount: 0
+featureList: [
+  "Drag-n-drop pointing elements and arrows to map out connectivity visually.",
+  "Document and explain design flows dynamically.",
+  "Easily define directions and pathways in visual templates."
+]
+howItWorks: "Collaborative Infrastructure as Design"
```
hiyach28, line 122: howItWorks appears to have been copied from meshery-core during the formatting update. Could we keep the Meshery Arrows-specific value ("Visual Flow Annotation") here? This metadata is model-specific, and "Collaborative Infrastructure as Design" describes Meshery Core rather than Meshery Arrows.

hiyach28 (approved):

collections/_models/meshery-arrows/meshery-arrows.md
dhruveshmishra: Thanks @hiyach28 ! I think we used Collaborative Infrastructure as Design to keep it consistent with mesherycore and over 200+ other models in the repo that use this exact standard heading. The arrow-specific details are already covered in the description and featureList.

banana-three-join (approved):
Ty for sticking out with one. Really appreciate it!

PR #2762, opened 2026-06-26
fix: initialize catalog modal content for dynamically loaded scripts

### LINKED ISSUE #2761: [Catalog] Fix DOMContentLoaded lifecycle issue in dynamically loaded catalog modal scripts
### Description

`catalog-modal.js` relies on the `DOMContentLoaded` event for initialization. However, this script is dynamically loaded after the page has already finished loading through `_includes/modal.html`.

As a result, the `DOMContentLoaded` callback never executes, preventing the "Read More" initialization logic from running for dynamically loaded modal content.

### Expected Behavior

The initialization logic should execute regardless of whether the script is loaded before or after the `DOMContentLoaded` event.

A possible solution is to use `document.readyState` and initialize immediately if the document has already loaded.

### Screenshots

N/A (JavaScript lifecycle issue)

### Environment:

* OS: macOS
* Browser: Chrome
* Version: Latest
* Device: Desktop

ananyaa0518:
## Description

This PR fixes https://github.com/meshery/meshery.io/issues/2761.

Previously, the script relied solely on the `DOMContentLoaded` event for initialization. Since `catalog-modal.js` is injected dynamically after the page has already loaded, the `DOMContentLoaded` event had already fired, preventing the initialization logic from executing.

This change introduces an `initializeReadMore()` function and uses `document.readyState` to determine whether to wait for `DOMContentLoaded` or initialize immediately.

### Changes

* Refactored initialization logic into `initializeReadMore()`
* Added `document.readyState` check
* Preserved existing behavior for scripts loaded before DOM completion
* Enabled proper initialization for dynamically injected modal scripts

### Testing

* Verified that modals continue to load correctly
* Verified that initialization executes when scripts are injected after page load
* Confirmed no console errors during modal interactions


ishwar170695:
This fixes the initialization issue. One thing I noticed while looking through the code is that `loadCatalogModalScript()` appends a new` <script>` element each time the modal is opened, so `catalog-modal.js` is executed repeatedly. It might be worth loading the script only once or guarding against duplicate inserts in a follow-up PR.

ananyaa0518:
Thank you for reviewing the PR and confirming the fix.

That's a good observation regarding loadCatalogModalScript(). I agree that repeatedly appending the script could lead to unnecessary re-execution of catalog-modal.js. I'll keep this in mind and can open a follow-up issue/PR to add a guard against duplicate script insertion or ensure the script is loaded only once.

Bhumikagarggg:
@ananyaa0518 Thank you for your contribution! Let's discuss this during the website call tomorrow at 5:30 PM IST | 7 AM CST Add it as an agenda item to the meeting minutes, if you would :slightly_smiling_face:

PR #2767, opened 2026-07-01
feat: add community leaderboard 

### LINKED ISSUE #947: Discussion Forum Integration: Create solutions leaderboard
**Current Behavior**
meshery.io does not integrate with its discussion forum (discuss.layer5.io). 

**Desired Behavior**
Ideally, one or more of the pages will display the a solutions leaderboard. 

See an example of this integration [in-action on this Meshery Docs page](https://docs.meshery.io/guides/mesheryctl). And [see this discussion](https://discuss.layer5.io/t/proposal-and-call-for-participation-building-the-community-leaderboard/1712/2) for more context.

See the [discuss.layer5.io design spec](https://docs.google.com/document/d/1MDdIZGr8cw_iyZD1H4G7k8AbcFJ79_4UhbKgGltfa7U/edit#heading=h.gqjd5dgruo7t) and the [moderating the community discussion forum docs](https://docs.google.com/document/d/1o4DydORiy4Cwe3UpbCg7TWGnt5ry7t-AXuddO2w0k0g/edit?pli=1#heading=h.mpw9xz2nzccu) for a list of other possibilities.

**Screenshots / Mockups**
Screenshot 2023-01-01 at 8 15 09 PM

---
<h4>Contributor [Guides](https://docs.meshery.io/project/contributing) and Resources</h4>

- 🎨 Wireframes and designs for Meshery Site in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🖥 [Contributing to Meshery Website](https://github.com/meshery/meshery.io#contributing-to-the-mesheryio-website)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](https://slack.meshery.io)

Maanvi212006:
**Description**
This PR fixes #947

Adds a community leaderboard to meshery.io that ranks members of the Layer5 discussion forum based on their contributions.

Inspired by Lee's PR [#6773 : [Docs] Integrate Discussion Forum with Posts by Tag](https://github.com/meshery/meshery/pull/6773) which introduced the static data + nightly CI pattern for integrating Discourse forum data into Meshery sites, this PR follows the same approach.

How it works:

- Nightly CI workflow — fetches data from the Discourse API, computes scores using the official formula (posts ×1 + likes ×2 + solutions ×3), and commits the result to `_data/leaderboard.json`
- Jekyll include — `_includes/leaderboard.html` reads the pre-computed data and renders a ranked table with avatars, usernames, and scores
- New page — accessible at `/community/leaderboard`

https://github.com/user-attachments/assets/6329839e-a5a6-4ce4-8c1e-e965a9448b51

Notes for Reviewers

- I checked. The workflow is running properly image
- The current approach uses static data refreshed nightly via CI, following the pattern from [meshery/meshery#6773](https://github.com/meshery/meshery/pull/6773). A live client-side fetch can be explored as a follow-up to increase data freshness.
- The fetch script now uses anonymous requests (no Api-Key/Api-Username headers), matching the pattern already used by `discuss/meshery.json`. No secret needs to be added post-merge.
- Data safety hardening (latest review round): the nightly script now refuses to overwrite `_data/leaderboard.json` if the `all` period comes back empty (weekly/monthly are left unguarded since they can legitimately be empty). On the frontend, the data script tag is parsed inside a try/catch with a safe fallback so a missing or malformed `leaderboard.json` degrades gracefully instead of throwing. The search filter also now shows a distinct "No members match that search." message when a query matches nothing.
- I plan to further improve this in follow-up PRs by adding pagination so the full user list is scored correctly (not just the top 50 by solutions), fuller response schema validation with fallback to the last-known-good data on API failure, and graceful fallbacks for missing avatars.

## Summary by CodeRabbit

* **New Features**
  * Added a Community Leaderboard with weekly, monthly, and all-time rankings.
  * Added period selection, username filtering, podium highlights, avatars, activity statistics, and responsive layouts.
  * Added leaderboard access to the Community navigation menu.
  * Leaderboard data refreshes automatically with updated ranking information.

* **Bug Fixes**
  * Added empty-state handling, fallback avatars, and safe display of user content.
  * Improved leaderboard presentation across desktop, tablet, mobile, and dark-mode views.


Maanvi212006:
Page exists on netlify preview not on github preview. I will fix that

banana-three-join:
@Maanvi212006 Tag me once that has been addresses so I can check it out

_includes/leaderboard.html
```diff
@@ -0,0 +1,108 @@
...
+        {% for user in site.data.leaderboard.users %}
+        {% if user.rank <= 3 %} <div class="podium-card podium-rank-{{ user.rank }}">
+            <div class="podium-card-body">
+                <div class="podium-avatar-wrapper">
+                    <span
+                        class="medal {% if user.rank == 1 %}medal-gold{% elsif user.rank == 2 %}medal-silver{% else %}medal-bronze{% endif %}">{{
+                        user.rank }}</span>
+                    <a href="{{ user.profile_url }}" target="_blank" rel="noopener">
+                        <img class="podium-avatar" src="{{ user.avatar }}" alt="{{ user.username }}"
+                            onerror="this.onerror=null; this.src='assets/images/nav-icons/community.svg';" />
```
Bhumikagarggg: Please use: this.src='/assets/images/nav-icons/community.svg';

_includes/leaderboard.html
```diff
@@ -0,0 +1,108 @@
...
+            </thead>
+            <tbody>
+                {% for user in site.data.leaderboard.users %}
+                {% if user.rank > 3 %}
+                <tr class="leaderboard-row {% cycle 'row-white', 'row-teal' %}">
+                    <td class="leaderboard-rank">{{ user.rank }}</td>
+                    <td class="leaderboard-member">
+                        <a href="{{ user.profile_url }}" target="_blank" rel="noopener">
+                            <img class="leaderboard-avatar" src="{{ user.avatar }}" alt="{{ user.username }}"
+                                onerror="this.onerror=null; this.src='assets/images/nav-icons/community.svg';" />
```
Bhumikagarggg: same. Add `/`

Bhumikagarggg (changes requested):
I tested the leaderboard update script locally with network access, and it works:

` node .github/scripts/fetch-leaderboard.js`

It saved 49 users and updated _data/leaderboard.json. In my run, only the last_updated timestamp changed, so the data fetch path seems functional.

However, I don’t think this is ready for approval yet. A few things need to be fixed:

1. git diff --check upstream/master...HEAD fails because of trailing whitespace / blank EOF issues in:
      - .github/workflows/update-leaderboard.yml
      - _data/navigation.yml
      - css/screen.scss
 
2. The new .github/scripts/fetch-leaderboard.js script does not pass the repo’s ESLint config because it uses Node globals like require, process, and __dirname, while the current ESLint config is browser-oriented. Please add a Node-specific ESLint override for .github/scripts/** or adjust the script/config so lint passes.

3. In _sass/leaderboard.scss, some styles look redundant or invalid:
      - box-shadow: #000000;
      - box-shadow: #000000, #000000;
      - border: #000000;
      - several dark-mode overrides repeat the same values as the default styles.
      - .leaderboard-filter sets color: #7A848E, which makes typed text grey even though the placeholder color is already set separately.

Bhumikagarggg (changes requested):
Image

recognition and leaderboard using same svg?

Maanvi212006:
> Image
> recognition and leaderboard using same svg?

@Bhumikagarggg  yes as I mentioned in the previous PR, I thought that this was matching with that 'recognition' icon. If it needs to be changed, let me know! And, I will update the PR as per the feedback provided

_includes/leaderboard.html
Maanvi212006: addressed

_includes/leaderboard.html
Maanvi212006: addressed

Bhumikagarggg:
> > Image
> > recognition and leaderboard using same svg?
> 
> @Bhumikagarggg yes as I mentioned in the previous PR, I thought that this was matching with that 'recognition' icon. If it needs to be changed, let me know! And, I will update the PR as per the feedback provided

Yes @Maanvi212006! Please change the icon 
example like : [Icon](https://www.google.com/url?sa=t&source=web&rct=j&url=https%3A%2F%2Fwww.vecteezy.com%2Fvector-art%2F53489040-leaderboard-icon-simple-design&ved=0CBYQjRxqFwoTCPiD3_rHuZUDFQAAAAAdAAAAABBq&opi=89978449)

Bhumikagarggg:
> I tested the leaderboard update script locally with network access, and it works:
> 
> ` node .github/scripts/fetch-leaderboard.js`
> 
> It saved 49 users and updated _data/leaderboard.json. In my run, only the last_updated timestamp changed, so the data fetch path seems functional.
> 
> However, I don’t think this is ready for approval yet. A few things need to be fixed:
> 
> 1. git diff --check upstream/master...HEAD fails because of trailing whitespace / blank EOF issues in:
>    
>    * .github/workflows/update-leaderboard.yml
>    * _data/navigation.yml
>    * css/screen.scss
> 2. The new .github/scripts/fetch-leaderboard.js script does not pass the repo’s ESLint config because it uses Node globals like require, process, and __dirname, while the current ESLint config is browser-oriented. Please add a Node-specific ESLint override for .github/scripts/** or adjust the script/config so lint passes.
> 3. In _sass/leaderboard.scss, some styles look redundant or invalid:
>    
>    * box-shadow: #000000;
>    * box-shadow: #000000, #000000;
>    * border: #000000;
>    * several dark-mode overrides repeat the same values as the default styles.
>    * .leaderboard-filter sets color: #7A848E, which makes typed text grey even though the placeholder color is already set separately.

@Maanvi212006 Please reply this also did you address this also?

Maanvi212006:
> > I tested the leaderboard update script locally with network access, and it works:
> 
> > 
> 
> > ` node .github/scripts/fetch-leaderboard.js`
> 
> > 
> 
> > It saved 49 users and updated _data/leaderboard.json. In my run, only the last_updated timestamp changed, so the data fetch path seems functional.
> 
> > 
> 
> > However, I don’t think this is ready for approval yet. A few things need to be fixed:
> 
> > 
> 
> > 1. git diff --check upstream/master...HEAD fails because of trailing whitespace / blank EOF issues in:
> 
> >    
> 
> >    * .github/workflows/update-leaderboard.yml
> 
> >    * _data/navigation.yml
> 
> >    * css/screen.scss
> 
> > 2. The new .github/scripts/fetch-leaderboard.js script does not pass the repo’s ESLint config because it uses Node globals like require, process, and __dirname, while the current ESLint config is browser-oriented. Please add a Node-specific ESLint override for .github/scripts/** or adjust the script/config so lint passes.
> 
> > 3. In _sass/leaderboard.scss, some styles look redundant or invalid:
> 
> >    
> 
> >    * box-shadow: #000000;
> 
> >    * box-shadow: #000000, #000000;
> 
> >    * border: #000000;
> 
> >    * several dark-mode overrides repeat the same values as the default styles.
> 
> >    * .leaderboard-filter sets color: #7A848E, which makes typed text grey even though the placeholder color is already set separately.
> 
> 
> 
> @Maanvi212006 Please reply this also did you address this also?
> 
> 

Yes, these are addressed

Bhumikagarggg:
@Maanvi212006 Thank you for your contribution! Let's discuss this during the website call tomorrow at 5:30 PM IST | 7 AM CST Add it as an agenda item to the meeting minutes, if you would :slightly_smiling_face:

Bhumikagarggg (dismissed):

leaderboard.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: page
+title: Community Leaderboard
+permalink: /community/leaderboard
+description: "Recognizing the most active members of the Layer5 discussion forum."
+---
+
+Celebrating the top contributors on the [Layer5 discussion forum](https://discuss.layer5.io). Rankings are based on posts created, likes received, and solutions accepted.
```
rishiraj38: Please update this to reference the Meshery discussion forum instead of the Layer5 discussion forum.

Replace `https://discuss.layer5.io` with `https://discuss.meshery.io`.

.github/scripts/fetch-leaderboard.js
```diff
@@ -0,0 +1,73 @@
+const fs = require('fs');
+const path = require('path');
+
+const API_KEY = process.env.DISCOURSE_API_KEY;
+const API_USERNAME = process.env.DISCOURSE_API_USERNAME || 'system';
+const API_URL = 'https://discuss.layer5.io/directory_items.json?period=all&order=solutions&limit=50';
```
rishiraj38: Change the API endpoint to use `discuss.meshery.io` instead of `discuss.layer5.io` to keep all network requests and references branded consistently with the Meshery site.

.github/workflows/update-leaderboard.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Run scoring script
+        run: node .github/scripts/fetch-leaderboard.js
+        env:
+          DISCOURSE_API_KEY: ${{ secrets.DISCOURSE_API_KEY }}
+
+      - name: Commit and push JSON
```
rishiraj38: Using a raw `git push` here might fail if the `master` branch has protection rules enabled (requiring PRs or status checks). 

In other workflows (like `discussions-data-files-update.yml`), we use `token: ${{ secrets.GH_ACCESS_TOKEN }}` in the checkout step and utilize the `stefanzweifel/git-auto-commit-action` action to safely pull, commit, and push back. It would be great to align this with that pattern to prevent potential runner failures.

leaderboard.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: page
+title: Community Leaderboard
+permalink: /community/leaderboard
+description: "Recognizing the most active members of the Layer5 discussion forum."
```
rishiraj38: replace `layer5` - `meshery`.

.github/scripts/fetch-leaderboard.js
```diff
@@ -0,0 +1,73 @@
...
+
+function buildLeaderboard(items) {
+    return (items || [])
+        .filter(item => item && item.user && item.user.username && !item.user.username.startsWith('anon'))
+        .map(item => ({
+            rank: 0,
+            username: item.user.username,
+            name: item.user.name || '',
+            avatar: item.user.avatar_template ? `https://discuss.layer5.io${item.user.avatar_template.replace('{size}', '200')}` : '',
+            profile_url: `https://discuss.layer5.io/u/${item.user.username}`,
```
rishiraj38: Same here. Please use `discuss.meshery.io` instead of `discuss.layer5.io`.

.github/scripts/fetch-leaderboard.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+function saveJSON(leaderboard) {
+    const output = {
+        last_updated: new Date().toISOString(),
+        users: leaderboard
+    };
+
+    const outputPath = path.join(__dirname, '../../_data/leaderboard.json');
+    fs.writeFileSync(outputPath, JSON.stringify(output, null, 2));
```
rishiraj38, line 78: To make this script robust if run in environments where the destination _data directory has not been checked out or created yet, we should ensure the parent directory exists before writing to it.

rishiraj38 (dismissed):

leecalcote (dismissed):
Please ensure that this is not for one and only one time period. 

Additionally, the default time period should not be "all-time".

Bhumikagarggg:
@Maanvi212006 Thank you for your contribution! Let's discuss this during the website call today (20 July) at 5:30 PM IST | 7 AM CST Add it as an agenda item to the [meeting minutes](https://meet.layer5.io/websites-minutes), if you would :slightly_smiling_face:

Maanvi212006:
Also found an unrelated CI failure on `.github/workflows/jekyll-pr-preview.yml` in the base repo — it's a `pull_request_target` workflow that tries to check out fork PR code directly, which GitHub blocks by default (pwn-request risk, since that trigger has access to secrets). Not something I can fix from my branch since it's outside this PR's files. Recommend splitting it into an unprivileged build job (on `pull_request`) that uploads an artifact, and a privileged deploy job (`workflow_run`) that only consumes the artifact. Happy to put together the fix if useful — let me know how you'd like to handle it.

leaderboard.md
Maanvi212006: addressed

.github/scripts/fetch-leaderboard.js
Maanvi212006: addressed

.github/workflows/update-leaderboard.yml
Maanvi212006: addressed

leaderboard.md
Maanvi212006: addressed

.github/scripts/fetch-leaderboard.js
Maanvi212006: addressed

.github/scripts/fetch-leaderboard.js
Maanvi212006: addressed

banana-three-join:
@Maanvi212006 That failing workflow will also part of the current workflow work being done so no worries about it

_includes/leaderboard.html
```diff
@@ -0,0 +1,173 @@
+<section class="leaderboard-section">
+
+    <div class="leaderboard-controls">
+        <div class="period-toggle" role="tablist" aria-label="Leaderboard time period">
+            <select id="period-select" class="period-select" aria-label="Leaderboard time period">
```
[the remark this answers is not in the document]
Maanvi212006: Accepted. tablist requires child role="tab" elements, but the container owns a `<select>`. Screen readers will announce this incorrectly.

_sass/leaderboard.scss
```diff
@@ -0,0 +1,500 @@
...
+        background-color 0.25s ease;
+}
+
+.leaderboard-filter::placeholder {
+    color: #7A848E;
+}
+
+.leaderboard-filter:focus {
+    border-color: #00B39F;
+}
```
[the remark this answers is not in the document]
Maanvi212006: Accepted. `outline: none` with no replacement breaks keyboard-only navigation — a WCAG issue

_sass/leaderboard.scss
```diff
@@ -0,0 +1,542 @@
...
+.podium-username {
+    font-size: 2.2rem;
+    font-weight: 700;
+    line-height: 1.1;
+    text-align: center;
+    color: #00B39F;
+    transition: color 0.2s ease;
+    width: 100%;
+    overflow-wrap: break-word;
+    word-break: break-word;
```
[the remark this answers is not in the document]
Maanvi212006: Accepted. `overflow-wrap: break-word` alone is sufficient in modern browsers; `word-break: break-word` is a legacy fallback that can cause unwanted mid-word breaks even when unnecessary.

.github/scripts/fetch-leaderboard.js
```diff
@@ -0,0 +1,95 @@
...
+const PERIODS = ['weekly', 'monthly', 'all'];
+
+async function fetchUsers(period) {
+    const headers = {};
+    if (API_KEY) {
+        headers['Api-Key'] = API_KEY;
+        headers['Api-Username'] = API_USERNAME;
+    }
+
+    const url = `${BASE_URL}?period=${period}&order=solutions&limit=50`;
```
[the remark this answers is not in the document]
Maanvi212006: Ranking by solutions before scoring means high posts/likes users could get excluded from the top-50 pool entirely. Pagination is a bigger change (needs to walk the full Discourse directory), so this could be tracked as a follow-up rather than block this PR on it.

.github/scripts/fetch-leaderboard.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    const data = await response.json();
+    return data.directory_items;
+}
+
+function computeScore(item) {
+    return ((item.post_count || 0) * 1) +
+        ((item.likes_received || 0) * 2) +
+        ((item.solutions || 0) * 3);
```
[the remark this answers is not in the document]
Maanvi212006: The specific symptom here (50 zero-score users) was actually caused by the auth issue — fixed now, weekly/monthly are returning real non-zero data after switching to anonymous requests. That said, the underlying suggestion (guard against a malformed/empty API response overwriting good data) is still good practice for a nightly cron job. Keeping it as a hardening follow-up, not urgent for this PR.

.github/workflows/update-leaderboard.yml
```diff
@@ -0,0 +1,42 @@
+name: Update Leaderboard
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
```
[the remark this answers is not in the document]
Maanvi212006: Accepted. Prevents a real race condition (two runs both pulling/writing the same file).

_sass/leaderboard.scss
```diff
@@ -0,0 +1,517 @@
...
+    .leaderboard-filter,
+    .period-select {
+        flex: 1 1 0;
+        width: auto;
+        max-width: none;
+        min-width: 0;
+        margin: 0;
+        padding: 0.6rem 0.5rem;
+        font-size: 0.78rem;
+    }
```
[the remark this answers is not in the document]
Maanvi212006: Accepted. Fixed both issues: moved the base `.period-select` block above the media query, and switched the mobile flex sizing from `.period-select` to its actual sibling `.period-toggle` so it now sizes correctly against `.leaderboard-filter`.

leaderboard.md
banana-three-join, line 1: **The permalink doesn't follow the repo convention**
 
`leaderboard.md` uses:
 
```yaml
permalink: /community/leaderboard
```
 
Its two siblings use the trailing-slash form:
 
```yaml
# recognition.md
permalink: /community/recognition/
redirect_from:
  - /community/badges
  - /community/recognition
 
# certifications.md
permalink: /community/certifications/
```
 
Master commit `48ed541a` ("Adopt trailing-slash permalink convention to fix redirect loops") established this deliberately: leaf pages get `permalink: /path/` plus `redirect_from: /path`, because the slashless permalink was causing redirect loops and 404s. Your branch already merged that commit in via `209a17d8`, so this page is now the only one in the repo on the retired pattern.
 
This is very likely the cause of your own note from Jul 2 — *"Page exists on netlify preview not on github preview."* Netlify's pretty-URL handling hides it; GitHub Pages doesn't.
 
```yaml
permalink: /community/leaderboard/
redirect_from:
  - /community/leaderboard
```
 
The nav entry can stay slashless — `redirect_from` covers it, the same way Recognition links to `/community/badges`.

eslint.config.mjs
banana-three-join, line 1: **The ESLint change disables linting rather than configuring it**
 
The new `.github/scripts/**` block supplies `languageOptions` but no `rules`, and the block that carries `js.configs.recommended` plus the custom rules now has `ignores: [".github/scripts/**/*.js"]`. Net effect: the script matches a config object with nothing in it, so no rule ever runs. Reproducible:
 
```
$ npx eslint .github/scripts/fetch-leaderboard.js
rc=0
$ printf '\nvar unusedThing = 1;\nif(true){bogusUndefinedFn()}\n' >> .github/scripts/fetch-leaderboard.js
$ npx eslint .github/scripts/fetch-leaderboard.js
rc=0
```
 
An undefined function call and an unused variable both pass. Flat config merges cumulatively, so a `languageOptions`-only override placed **after** the rules block gets you Node globals while keeping every rule:
 
```js
{
  files: [".github/scripts/**/*.js"],
  languageOptions: {
    sourceType: "commonjs",
    globals: { ...globals.node },
  },
}
```
 
Delete the `ignores: [".github/scripts/**/*.js"]` line from the browser block. With that applied, ESLint resolves `require` / `process` / `__dirname` cleanly and reports 55 `indent` errors — all `--fix`-able in one pass, since the script is 4-space indented while the repo standard is 2.
 
Separately: roughly 130 lines of the `eslint.config.mjs` diff are a whole-file 4→2 space re-indent unrelated to this PR. Please revert the reformatting so the config change is reviewable — keep the added trailing newline.

.github/workflows/update-leaderboard.yml
```diff
@@ -0,0 +1,48 @@
...
+        uses: actions/checkout@v6
+        with:
+          token: ${{ secrets.GH_ACCESS_TOKEN }}
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Install dependencies
```
banana-three-join: `npm install undici` before `git pull` will break the job intermittently

npm ≥7 saves by default, so that step writes a `dependencies` block into `package.json` and rewrites `package-lock.json`, leaving a modified working tree:
 
```
$ npm install undici
$ git status --porcelain
 M package-lock.json
 M package.json
```
 
The next step is `git pull origin master`, and git refuses to merge when incoming commits touch a locally-modified file — *"Your local changes to the following files would be overwritten by merge: package.json."* It'll fail on any night master carries a dependency bump.
 
Node 20 has global `fetch`, so the cleanest fix is to drop `undici` entirely: remove the `require('undici')` line from the script and delete the whole "Install dependencies" step.

.github/workflows/update-leaderboard.yml
```diff
@@ -0,0 +1,48 @@
+name: Leaderboard data update
+
+concurrency:
+  group: leaderboard-update
+  cancel-in-progress: false
+
+on:
+  workflow_dispatch:
+  schedule:
```
banana-three-join, line 9: Cron collides with the existing discussions workflow
 
`discussions-data-files-update.yml` is also on `'0 0 * * *'`, and both workflows `git pull` and push to `master`. Your `concurrency` group only guards this workflow against itself, so expect intermittent rejected pushes. Stagger it:
 
```yaml
schedule:
  - cron: '30 1 * * *'
```

.github/scripts/fetch-leaderboard.js
```diff
@@ -0,0 +1,91 @@
...
+const BASE_URL = 'https://discuss.meshery.io/directory_items.json';
+const PERIODS = ['weekly', 'monthly', 'all'];
+
+async function fetchUsers(period) {
+    const headers = {
+        'User-Agent': 'meshery-leaderboard-bot/1.0',
+        'Accept': 'application/json'
+    };
+
+    const url = `${BASE_URL}?period=${period}&order=likes_received&limit=50`;
```
banana-three-join: Ranking pool is biased by the fetch ordering
 
```js
`${BASE_URL}?period=${period}&order=likes_received&limit=50`
```
 
The candidate pool is the top 50 *by likes received*, then re-ranked by a score that weights likes 2×. That's self-reinforcing: a member with many posts and accepted solutions but modest likes never enters the pool, so they can't place regardless of score. For an issue titled *"Create solutions leaderboard"* that's a meaningful gap.

leaderboard.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: page
+title: Community Leaderboard
+permalink: /community/leaderboard
+description: "Recognizing the most active members of the Meshery discussion forum."
+---
+
+Celebrating the top contributors on the [Meshery discussion forum](https://discuss.meshery.io). Rankings are based on posts created, likes received, and solutions accepted.
```
banana-three-join: Maybe the wording could be enhanced a little bit now that the default is the weekly leaderboard instead of the all time one.

_sass/leaderboard.scss
banana-three-join, line 1: Could you check the following findings from my search across the current rules?

**Likely unintended:**
 
- `.medal { border: 4px solid; }` has no colour, so it resolves to `currentColor`. `medalClass()` returns gold, silver, or bronze for every rank and all three set `color: #000000`, so every medal gets a black ring — while `.medal`'s own `color: #ffffff` is dead. If the intent was a white separator ring, set it explicitly: `border: 4px solid #ffffff`.

**Redundant — safe to delete:**
 
- `.leaderboard-rank { color: #000000 }` (L302) is inert: `.leaderboard-table td` has higher specificity (0-1-1 vs 0-1-0), matches the same elements, and sets the same value.
- `.row-teal .leaderboard-member a:hover .leaderboard-username` (L337–340) has declarations identical to the general rule directly above it at L332.
- `.podium-rank-3 { order: 3 }` inside the 600px query duplicates the base rule at L87 with the same value.
- `.leaderboard-podium` is declared twice **inside the same 600px query** (L468 and L537), with `align-items: center` in both. Merge them.
- `.podium-card { background: transparent; border: none; box-shadow: none }` — nothing else in the codebase targets `.podium-card` (the `.card` rules in `catalog.scss` / `handbook.scss` match exact class names, not substrings), so these three reset nothing.
- `image-rendering: auto` (L113) is the CSS initial value.
- `border-color: #00B39F` in `.leaderboard-filter:focus-visible`, `.period-select:focus-visible`, and `body.dark-mode .period-select` all restate the base border colour — only the `box-shadow` focus rings do any work.

**Consistency:**
 
- The file hardcodes hex throughout, while `rootvariables.scss` already provides `--brand-color-secondary` and friends that flip automatically under `.dark-mode`. Adopting them would let you delete most of the hand-written `body.dark-mode` blocks here. (To be clear, the `.dark-mode` `border-color` and `td { color: #000000 }` overrides are *necessary* — the global `table, th, td { border: 1px solid var(--color-secondary-dark) }` inverts to near-white in dark mode against the hardcoded white rows. They're doing real work.)
- Dark-mode selectors are mixed: `body.dark-mode` in seven places, a bare nested `.dark-mode` at L314.
- Media queries are interleaved with base rules — the 768px block sits at L365, then base `.period-select` rules follow at L391. No conflict today, but any future base rule for a class styled in that block will silently win over it. Moving both queries to the end of the file avoids the trap.

_includes/leaderboard.html
banana-three-join, line 1: Please address the following findings:

- **Zebra striping breaks under filtering.** `row-white` / `row-teal` are assigned from the render-time index, so filtering leaves runs of same-coloured rows. Unfiltered: `row-white, row-teal, row-white, row-teal, row-white`. Filtered on `"a"`: `row-white, row-white, row-teal, row-white`. Move it to CSS with `tbody tr:nth-child(odd)`, or reassign on filter.
- **Prettier has mangled the Liquid tag.** `{{ site.data.leaderboard.default_period | jsonify` and `}}` are split across lines, everything after it is dedented a level, and the IIFE closes as `}) ();`. Liquid's tokenizer is newline-tolerant so it should parse — but no one has built the current tip, so please confirm against a fresh preview and reformat with the Liquid tag kept on one line.
- **`onkeyup` + a `window` global.** `filterLeaderboard` is exported to `window` purely to serve the inline `onkeyup=` attribute. Binding `addEventListener('input', ...)` inside the IIFE keeps the closure private and also catches paste-by-mouse and autofill, which `keyup` misses.
- **Accessibility:** the search input has only a placeholder and no label; the `aria-label` on `.period-toggle` duplicates the one on the `<select>` it wraps; table headers want `scope="col"`.

.github/scripts/fetch-leaderboard.js
```diff
@@ -0,0 +1,91 @@
...
+        const body = await response.text();
+        throw new Error(`Discourse API error (period=${period}): ${response.status} - ${body.slice(0, 200)}`);
+    }
+
+    const data = await response.json();
+    return data.directory_items;
+}
+
+function computeScore(item) {
+    return ((item.post_count || 0) * 1) +
```
banana-three-join: ((item.post_count || 0) * 1)`** — the `* 1` is a no-op; you can drop it

.github/scripts/fetch-leaderboard.js
```diff
@@ -0,0 +1,91 @@
...
+const PERIODS = ['weekly', 'monthly', 'all'];
+
+async function fetchUsers(period) {
+    const headers = {
+        'User-Agent': 'meshery-leaderboard-bot/1.0',
+        'Accept': 'application/json'
+    };
+
+    const url = `${BASE_URL}?period=${period}&order=likes_received&limit=50`;
+    const response = await fetch(url, { headers });
```
banana-three-join: A timeout and retry could also be good. Not needed, but worth looking into to not burn the runner minutes in case of a hung request

_includes/leaderboard.html
banana-three-join, line 1: Could you please verify that `username`, `name`, `avatar`, and the numeric counts can't contain `"`, `<`, or `>` before they're interpolated into HTML attributes and the `<script type="application/json">` block — if any of them can, add `.replace(/"/g, '&quot;')` to `escapeHtml`, coerce the counts with `Number(...) || 0`, and drop the unused `name` field from the emitted JSON.

banana-three-join (commented):
Looking really good! Just a couple of items that I would like for you to address before approving and merging:

**Please guard against an empty response overwriting good data**
 
`buildLeaderboard` swallows a missing payload via `(items || [])`. If Discourse returns `200` with an empty body, the script writes empty arrays, the bot commits them, and the live page flips to "No activity for this period yet."
 
Don't guard on every period — `period=weekly` legitimately returns empty during a quiet week, and you'd get failing runs. `all` is the one that should never be empty on a live forum:
 
```js
if (!periods.all.length) {
  throw new Error('all-time period empty — refusing to overwrite');
}
```
 
**Handle a missing or reshaped data file**
 
If `_data/leaderboard.json` is absent or lacks `periods`, `{{ ... | jsonify }}` emits `null`, `JSON.parse` succeeds and returns `null`, and the next line throws:
 
```
Uncaught TypeError: Cannot read properties of null (reading 'weekly')
```
 
The section chrome renders with an empty podium, an empty table, and the empty-state paragraph still hidden — a dead panel with no message. Wrap the read and fall back to the empty state:
 
```js
var leaderboardData;
try {
  leaderboardData = JSON.parse(document.getElementById('leaderboard-data').textContent) || {};
} catch (e) {
  leaderboardData = {};
}
```
 
**Show the empty state when a search matches nothing**
 
Filtering on a name that doesn't exist hides every row and every podium card, but `#leaderboard-empty-state` stays at `display: none`. Verified: 0 visible rows, 0 visible cards, no message. The element already exists — reuse it with a distinct message like "No members match that search."

Also, please update the PR Description. It says `DISCOURSE_API_KEY` needs adding to repo secrets after merge, but commit `39e906ce` removed Discourse auth and the script sends no key. As written, someone will provision a secret that's never read.

_sass/leaderboard.scss
Maanvi212006: decided to leave the `#ffffff`/ `#000000` values as-is rather than swapping to `--color-primary-light`/`--color-secondary-light`.

Those tokens resolve to off-white/charcoal in this theme, not pure white/black. The leaderboard cards and table are intentionally pure black-on-white (light) and white-on-black (dark) — same reasoning you called out for why the `.dark-mode` border/text overrides on the table are necessary rather than redundant. Adopting the tokens here would visibly shift the colors, so keeping the hardcoded values is a deliberate choice, not an oversight.

Happy to revisit if off-white/charcoal is actually what's intended for the leaderboard design — let me know.

leaderboard.md
Maanvi212006: Accepted. The permalink must follow the rpeo convention.

eslint.config.mjs
Maanvi212006: Accepted. The `ignores` line on the browser block combined with the rules-less scripts block meant nothing was actually being linted for `.github/scripts/**`.

Fixed:

- Removed the `ignores: [".github/scripts/**/*.js"]` line from the browser config block.
- Added a `languageOptions`-only override for `.github/scripts/**/*.js` (Node globals, commonjs) placed after the rules block, so it merges cumulatively and keeps all the existing rules active instead of replacing them.
- Reverted the unrelated 4→2 space reformat of the file — the diff should now just be the scripts block addition and the one-line removal from the browser block.
- Ran `--fix` on `fetch-leaderboard.js`; the indent errors are resolved.

Thanks for catching this — happy to confirm `npx eslint .github/scripts/fetch-leaderboard.js` now runs clean.

.github/workflows/update-leaderboard.yml
Maanvi212006: Addressed

.github/workflows/update-leaderboard.yml
Maanvi212006: Addressed

leaderboard.md
Maanvi212006: Addressed

_includes/leaderboard.html
Maanvi212006: Addressed.

.github/scripts/fetch-leaderboard.js
Maanvi212006: Addressed

banana-three-join (commented):
Two small things came in with the changes that are visible on the page:

1. **`.visually-hidden` isn't defined anywhere in the repo** — that new label will render as visible text above the search box. Needs the clip pattern added to `leaderboard.scss`.
2. **Switching to a period with no activity, then typing in the search, blanks the section.** `renderLeaderboard` returns early without clearing the podium/tbody, so the stale rows make `filterLeaderboard` hide the empty state while the containers are still hidden — nothing renders, no message. Clearing both in that early-return branch fixes it.

A few leftovers:

- `.row-white` / `.row-teal` are now dead CSS since the JS stopped emitting those classes.
- The script has 57 `indent` errors now that linting works — one `--fix` run.
- `leaderboard.md` says "switch periods above", but the selector renders below that paragraph.
- `_data/leaderboard.json` still has `name` on every record; worth re-running the script and committing so it matches.
- The date still renders as a raw ISO string — `| date: "%B %-d, %Y"`.
- The ranking pool is still `order=likes_received&limit=50`, so members strong on posts/solutions but light on likes can't place regardless of score. That's the one substantive item left.

_includes/leaderboard.html
Maanvi212006: Yes — `username`/`avatar` come straight from the Discourse API with no guarantee on content, so treating them as untrusted is the right call. Applied all three:

- `escapeHtml` now also escapes `"` so a username or avatar URL containing a quote can't break out of the `href`/`src`/`alt` attributes it's interpolated into.
- `likes`, `posts`, `solutions`, and `score` are now coerced with `Number(...) || 0` at every render site (podium and table), so a malformed/non-numeric API response can't land raw in the DOM.
- Dropped the `name` field entirely — it was written into `leaderboard.json` but never read anywhere in `leaderboard.html`, so removing it shrinks the emitted JSON with no behavior change.

Maanvi212006:
Both flagged — agreed on both. I'd lean toward tracking them in a follow-up PR rather than expanding scope here, but happy to do either — let me know which you'd prefer.

- [Ranking pool bias](https://github.com/meshery/meshery.io/pull/2767#discussion_r3827508443) (top-50-by-likes before scoring): same root cause as the earlier top-50-by-solutions gap, needs pagination through the full directory before scoring rather than a single-metric-capped pool. 
- [Fetch timeout/retry](https://github.com/meshery/meshery.io/pull/2767#discussion_r3827582032): agreed this is good hygiene to avoid burning runner minutes on a hung request.

.github/scripts/fetch-leaderboard.js
Maanvi212006: Refer [here](https://github.com/meshery/meshery.io/pull/2767#issuecomment-5420700668) !

.github/scripts/fetch-leaderboard.js
Maanvi212006: Refer [here](https://github.com/meshery/meshery.io/pull/2767#issuecomment-5420700668) !

banana-three-join (approved):
@Maanvi212006 Let's go ahead and make both follow-up PRs to see how the leaderboard behaves beforehand, as these shouldn't block the PRs' mergeability.

PR #2779, opened 2026-07-08
[Site] Fix 404s on trailing-slash page URLs with redirect_from

### LINKED ISSUE #2777: [Bug] Trailing slash on page URLs causes broken/404 pages
### Description

Several pages on `meshery.io` break when accessed with a trailing slash, even though the non-slash version works correctly. For example:
- ✅ `https://meshery.io/blog` → works
- ❌ `https://meshery.io/blog/` → broken page
This happens because Jekyll generates these pages with a `permalink` that has no trailing slash, and there is no `redirect_from` entry to absorb the trailing-slash variant. GitHub Pages does not automatically redirect `/page/` → `/page`.
### Affected Pages (there can be more)
| Page | Canonical URL | Trailing slash handled? |
|------|--------------|------------------------|
| Blog index | `/blog` | ❌ Missing `redirect_from: /blog/` |
| Catalog | `/catalog` | ❌ Missing |
| Community | `/community` | ❌ Missing |

### Expected Behavior

Accessing any page with or without a trailing slash should resolve correctly, either serving the page directly or redirecting to the canonical URL.

### Fix
The `jekyll-redirect-from` plugin is already installed and used correctly on several pages (`/brand`, `/charts`, `/video`, `/thank-you`, `/service-mesh-interface`). The same pattern needs to be applied to the affected pages.
Example fix for `blog/index.html`:
```yaml
---
permalink: /blog
redirect_from:
  - /blog/
---
```

### Screenshots

Image

### Environment:
 - OS: Mac Linux Windows
 - Browser: Chrome Safari Firefox
 - Version: 
 - Device: Desktop Mobile

---

- 🙋🏾🙋🏼 Questions: [Discussion Forum](/community#discussion-forums) and [Community Slack](https://slack.meshery.io)

jfu06:
**Description**

This PR fixes #2777

Pages generated with an extensionless `permalink` (e.g. `/blog`) break when accessed with a trailing slash (`/blog/`), because GitHub Pages serves static files only: `/blog` falls back to `blog.html`, but `/blog/` looks for `blog/index.html`, which the build does not produce. This PR applies the existing `jekyll-redirect-from` pattern (already used on `/brand`, `/charts`, `/video`) so every affected page emits a redirect page at its trailing-slash variant.

Beyond the three pages named in the issue (`/blog`, `/catalog`, `/community`), the same defect existed on `/calendar`, `/subscribe`, `/extensions`, `/programs`, `/features`, `/newcomers`, `/talks-and-trainings`, `/blog/category`, the catalog subpages (`/catalog/designs`, `/catalog/models`, `/catalog/filters`, `/catalog/webassembly`), and the community subpages (`/community/newcomers`, `/community/certifications`, `/community/recognition`, `/community/meshmates`). All are fixed here.

Two special cases:

- **Blog index**: the issue's suggested fix (`redirect_from: /blog/` on `blog/index.html`) does not work there — octopress-paginate clones the page front matter onto every pagination page, so `/blog/` ends up redirecting to the last pagination page (`/page5`) instead of `/blog`. This PR instead adds a small stub page (`collections/_pages/blog-trailing-slash.html`) using `redirect_to`, which generates the correct redirect.
- `/extensions` and `/programs` already had this exact fix present but commented out; this PR enables it. The `/features` front matter had a `redirect_from` entry pointing at its own permalink, now corrected to the trailing-slash variant.

**Notes for Reviewers**

- This bug is not reproducible with `jekyll serve` / `make site`: the dev server has its own trailing-slash fallback that GitHub Pages does not have. Verify against the built `_site` output instead.
- Verified with a full local `jekyll build` (Docker, `jekyll/jekyll:4`): all 19 trailing-slash paths now emit a redirect page targeting their canonical URL, and the canonical pages plus blog pagination (`/page2`…`/page5`) are unaffected.
- The live `/blog/` today serves a stale, unrendered copy of the page source (the "broken page" in the issue screenshot) left on the `gh-pages` branch by `keep_files: true` in the deploy workflow. The redirect page generated by this PR deploys to the same path and overwrites it. The `keep_files: true` setting itself may deserve a separate issue, as it can preserve other stale artifacts indefinitely.


leecalcote:
Thank you, @jfu06.

@Bhumikagarggg, @banana-three-join, please assist @jfu06 in getting this to a final state.

banana-three-join:
I think this fix needs a different approach.

`redirect_from` creates a stub (an index.html file with a redirect to the page in which the `redirect_from` is declared). For example, let's say your permalink is `/path`, by setting the `redirect_from: /path/` directive, you create a `/path/index.html` file which redirects you to `/path`. This is needed for github pages due to how strict it is, if you add a trailing-slash /, it'll consider the path a directory and look for its index, but since it doesn't exist, you get the 404 error despite it working when served locally through Jekyll's server. The issue? Try running those changes locally, example: /blog/category with redirect_from: /blog/category/. You'll run into the following behavior:

image

Why? Because /blog/category/ is getting resolved into /blog/category/index.html (the stub) which redirects to /blog/category, but the server now sees that a /blog/category/ directory exists because of the stub, so it redirects back to /blog/category/, which becomes an endless loop.

I propose changing the permalink convention from /path to /path/, this creates a /path/index.html and for users who are used to /path, that's where we use the redirect_from so the original behavior is kept as they'll now be redirected to /path/. This would be only for leaf pages as directory landings already properly resolve with the change from /path to /path/.

Also, as a side note but also just as important, a lot of the pages have malformed links so that's something which could be cleaned up as we go as well. @jfu06

jfu06:
Thanks @banana-three-join, great catch — you're right, the stub approach loops. I've reworked the PR to invert the convention as you proposed: permalinks are now /path/, with redirect_from: /path for backwards compatibility on leaf pages only. Directory landings (/blog, /catalog, /community, etc.) just get the permalink change since their children already resolve the slashless form. I also applied the same fix to /brand, /charts, /thank-you, and /service-mesh-interface, which had already shipped the old pattern and loop in production today, and cleaned up the malformed redirect entries I found in passing (duplicate redirect_from key in smi.html, missing leading slashes in talks-and-trainings.html). One side effect worth noting: the paginated blog pages move from /page2 to /blog/page2. Verified with a local Jekyll build, happy to adjust further!

leecalcote:
I'm marking with high priority. This bug is a gnarly experience for the unwitting site visitor.

leecalcote:
@jfu06, regarding:

>  One side effect worth noting: the paginated blog pages move from /page2 to /blog/page2.

This sounds like a desirable side effect. 👍

banana-three-join:
@jfu06 Ty! Let me go ahead and test it on my end so we can get this closed.

banana-three-join:
**Notes for reviewers:**

**Collection item URLs — intentionally unchanged**

The trailing-slash convention was applied to landing and leaf pages, but **not** to collection item permalinks (`extensions`, `models`, `filters`, `custom-models`, `catalog`). These keep their existing no-slash / `.html` canonical form (e.g. `/extensions/github-action-performance-analysis`, `/catalog/resiliency/foo.html`).

Requesting the slashed variant (e.g. `/extensions/foo/`) returns a 404 in production, but this is harmless: a grep across templates, includes, and layouts confirms nothing links to the slashed form — the only way to hit it is by manually typing a trailing slash. Changing these would alter hundreds of canonical URLs to fix paths that are never requested, so the config is left as-is.

If slashed item URLs ever need to resolve, append `/` to the collection permalinks in `_config.yml` — but only for the extensionless collections, never `catalog` (a trailing slash on a `.html` URL is invalid).

banana-three-join (approved):

banana-three-join:
Should be good. Thanks a lot for the help! @jfu06

PR #2839, opened 2026-07-28
Fix: make site Docker startup for local Jekyll development

saurabhraghuvanshii:
**Description**

 PR fixes the `make site` Docker workflow for local development.

Previously, `make site` started the Jekyll container in detached mode and returned only a container ID. If the container failed during startup, the failure was hidden because the container exited and was removed immediately. The actual failure was caused by missing Gemfile dependencies inside the `jekyll/jekyll:4` image.

### This update makes the Docker target:

- run Jekyll in the foreground so startup errors are visible
- install missing bundle dependencies inside the container
- cache gems in a named Docker volume
- bind Jekyll to `0.0.0.0` so the site is reachable from the host
- expose both the site port and LiveReload port

This PR fixes #

**Notes for Reviewers**

## Summary by CodeRabbit

* **Chores**
  * Improved the Docker-based development workflow by automatically installing missing gems before starting the site.
  * Persisted installed gems between container runs for faster startup.
  * Updated the development server to be accessible from outside the container.
  * Ensured cleanup removes the development container reliably.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,9 +14,8 @@ build:
 	$(jekyll) build --drafts
 
 docker:
-	@# Run jekyll in official jekyll image without running bundle install
-	@# This avoids fetching gems on the host when Docker provides a working jekyll runtime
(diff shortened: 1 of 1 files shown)


Sbragul26 (approved):
LGTM :+1:

rishiraj38 (approved):

PR #2847, opened 2026-07-31
[chore]: Split PR preview into build/deploy workflows; drop `pull_request_target`

banana-three-join:
## Summary

Reworks the PR preview pipeline so that untrusted pull request code is never
executed in a job that holds write credentials. The single `pull_request_target`
workflow is replaced by an unprivileged build workflow and a privileged
`workflow_run` deploy workflow that treats the build output as data.

### Trust boundary

The old workflow checked out and built fork PR code in a job with
`contents: write` and repository secrets. `actions/checkout` v7 now rejects that
pattern outright, and the previous workflow only continued to function because of
`allow-unsafe-pr-checkout: true`. Since the enforcement is being backported to all
supported majors, removing the flag rather than carrying it forward was the only
durable option.

Build now runs on `pull_request` with `contents: read` and
`persist-credentials: false`, producing `_site/` and PR metadata as an artifact.
Deploy runs on `workflow_run` and fetches that artifact by `run-id` — it never
checks out PR code, so nothing the contributor authored executes with the write
token.

This makes the artifact the trust boundary, which means its contents are now
untrusted input. The PR number and event action are therefore validated on the
privileged side before use; a non-numeric number or an unrecognised action fails
the job. Without this, a crafted artifact could direct the deploy at an arbitrary
PR or invert the deploy/remove decision.

## Summary by CodeRabbit

* **New Features**
  * Added automated documentation previews for pull requests.
  * Preview links are posted automatically when builds complete.
  * Preview sites are removed when pull requests close.
  * Old previews are automatically pruned to maintain a six-preview limit.
  * Preview pages discourage search engine indexing.

* **Chores**
  * Replaced the previous pull-request preview workflow with separate build and deployment workflows.
  * Improved preview deployment reliability and validation.


.github/workflows/build-pr-preview.yml
lekaf974, line 1: this file indentation does not match 2 spaces for yaml file

banana-three-join: Indentation has been fixed.

PR #2868, opened 2026-08-10
Add category sections and side navigator to extensions page

### LINKED ISSUE #2814: [Website]: Improve the intra-page section navigation for /meshery.io/extensions
### Current Behavior
The current extensions page does not have an intra-section bifurcation for different categories of extensions.

### Desired Behavior
Add a slight distinction (an `<hr>` separator with a small label-like tag) between each category of extensions on https://meshery.io/extensions.

The behaviour should support an intra-page section navigator to scroll the visitor to that section when clicked.

### Screenshots / Mockups

**Current Setup**
Image

**Mockup**

Image

---
<h4>Contributor [Guides](https://docs.meshery.io/project/contributing) and Resources</h4>

- 🖥 [Contributing to Meshery Website](https://github.com/meshery/meshery.io#contributing-to-the-mesheryio-website)

Himanshujha7:
**Description**
Group extensions by category under an hr divider tagged with the
category chip, and add a fixed right-hand navigator that scrolls to
each section. Sorting now applies within a category; search hides
categories left empty.

This PR fixes #2814 

**Notes for Reviewers**

Screenshots/Videos:

image

https://github.com/user-attachments/assets/5509d135-ed4f-4d8d-8a67-e49195a13d28

## Summary by CodeRabbit

* **New Features**
  * Added extension catalog filtering with category checkboxes, result counts, and independent sorting by category.
  * Added responsive filter controls with mobile-friendly toggle, close, and interaction states.
  * Added category dividers and improved extension catalog navigation.
  * Search and category filters now work together, including a global no-results message.
* **Updates**
  * Updated the Kanvas extension documentation link.
  * Refined `PERFORMANCE` chip branding for improved visual consistency.


collections/_pages/extensions.html
```diff
@@ -1,188 +1,355 @@
...
+{% comment %}
+  Keep only the categories that have something to list, so the navigator never
+  links to an empty section.
+{% endcomment %}
+{% assign categories = "" | split: "" %}
+{% for type in all_types %}
+  {% assign type_count = 0 %}
+  {% for extension in site.extensions %}
+    {% if extension.type == type %}{% if extension.extensionInfo or extension.patternInfo and extension.Status !=
+    "ComingSoon" %}
```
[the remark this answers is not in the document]
Himanshujha7: Thank you for the review!
I have made the necessary fix for `coming-soon` and have pushed the changes in latest commit.

_sass/catalog.scss
```diff
@@ -1,1270 +1,1413 @@
...
+      box-shadow: none;
+      background-color: transparent;
+      color: var(--color-secondary-medium);
+    }
+  }
+
+/* EXTENSION CATEGORY SECTIONS */
+.extension-category {
+  width: 100%;
+  // Clears the sticky site header.
```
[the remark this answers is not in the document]
Himanshujha7: Hi,
I have made the necessary changes in the latest commit

Himanshujha7:
Hi @banana-three-join 
Kindly review and let me fir any further changes are needed for this PR.

collections/_pages/extensions.html
```diff
@@ -55,121 +55,195 @@
...
+      }
+
+      if (matches) {
         found = true;
-      } else {
-        extension.style.display = "none";
       }
     });
 
+    categoryNav.style.display = found ? "" : "none";
```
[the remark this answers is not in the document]
Himanshujha7: This has been addressed in latest commit.

collections/_pages/extensions.html
```diff
@@ -55,121 +55,195 @@
...
+          if (a.getAttribute("status") === "coming-soon") {
+            return 1;
+          } else if (b.getAttribute("status") === "coming-soon") {
+            return -1;
+          } else {
+            return b
+              .getAttribute("filter")
+              .localeCompare(a.getAttribute("filter"));
+          }
+        });
```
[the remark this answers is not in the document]
Himanshujha7: This has been addressed in latest commit. Please review it

_sass/catalog.scss
```diff
@@ -340,8 +340,154 @@ p {
...
+$category-nav-width: 7.5rem;
+$category-nav-tick: 1.25rem;
+
+.category-nav {
+  position: fixed;
+  right: 1.5rem;
+  top: 50%;
+  transform: translateY(-50%);
+  // Below the site header ($z-index-header) and the card modals.
+  z-index: 5;
```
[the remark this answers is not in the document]
Himanshujha7: Please review the latest commit.

dhruveshmishra (commented):
Image

I think this overlapping is giving somewhat of odd experience, u can do one thing that on scrolling down the icons comes out from the sections, and on scrolling up it the name should be merged into the main category name to a single. Whats about this?

Himanshujha7:
@dhruveshmishra 
That's a very good point-out.
Strange that I didn't see the overlap anyway; let me quickly check this.
Also, just curious: will that not be too complex a UI at one particular section of the page?
If it is fine to do that, I can think of a way to do it.

dhruveshmishra:
yah! u can try a new approach if its there in your mind, we can take a look on that how its work.

Maanvi212006 (changes requested):
The sidebar navigation menu is not visible in mobile view. 

Image

Maanvi212006:
> @dhruveshmishra That's a very good point-out. Strange that I didn't see the overlap anyway; let me quickly check this. Also, just curious: will that not be too complex a UI at one particular section of the page? If it is fine to do that, I can think of a way to do it.

Regarding this — do you have an approach in mind?
From my side, one option could be to lay the category tags out horizontally across the top of each card's section, so they span the full width instead of stacking on top of each other. That way there's no overlap between the tag and the card content underneath.

Himanshujha7:
> > @dhruveshmishra That's a very good point-out. Strange that I didn't see the overlap anyway; let me quickly check this. Also, just curious: will that not be too complex a UI at one particular section of the page? If it is fine to do that, I can think of a way to do it.
> 
> Regarding this — do you have an approach in mind? From my side, one option could be to lay the category tags out horizontally across the top of each card's section, so they span the full width instead of stacking on top of each other. That way there's no overlap between the tag and the card content underneath.

Hi @Maanvi212006 Do you mean as an accordian having toggle features?
Since the initial idea was to create a navigator for the sections - I thought of this way as of now.

I am just thinking of making it more polished having the current implementation.
Please let me know your perspective.

Also, regarding the mobie view - Thanks for pointing it out, I will try to implement this as well in the latest commit.

bin/jekyll.cmd
```diff
@@ -0,0 +1,19 @@
+@ruby -x "%~f0" %*
+@exit /b %ERRORLEVEL%
```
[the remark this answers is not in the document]
Himanshujha7: This was done by me to run the site locally which I have reverted back in the latest commit

Makefile
```diff
@@ -4,22 +4,24 @@ ARTIFACTHUB_SCRIPT_DIR := $(REPO_ROOT)/assets/scripts
...
 	@# Run Jekyll with the repo's Gemfile dependencies isolated in a Docker volume.
 	docker run --name meshery-io --rm -p 4000:4000 -p 35729:35729 -v "$(CURDIR)":"/srv/jekyll" -v meshery-io-bundle:"/usr/local/bundle" -w /srv/jekyll jekyll/jekyll:4 sh -lc 'bundle check || bundle install; bundle exec jekyll serve --host 0.0.0.0 --drafts --livereload --config _config_dev.yml'
-
 docker-stop:
 	docker stop meshery-io
-
+	rm -rf .jekyll-metadata
+	rm -rf .jekyll-cache
+	docker stop meshery-io || true
+	docker rm meshery-io || true
```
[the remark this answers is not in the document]
Himanshujha7: Hi, Thanks for pointing it out.
I have removed the changes since this was done to test the site locally but I have removed them in the latest commit now.

sangramrath:
Why not implement a similar UI/UX we have for https://meshery.io/catalog/models/, i.e check boxes on left for categories. Unless that's also something we are planning to change.

Screenshot 2026-08-17 at 6 19 25 PM

Himanshujha7:
Hi @sangramrath 
Yes! Thank you for the suggestion in the meeting as well. Lenox has raised another issue, as per your suggestion from Monday's website meeting, which I will work on to add similar vertical navigation.

To keep the scope clear for now, I'll implement the meeting's suggestion for this PR and submit a separate PR covering the above scope.

Katotodan (commented):
@Himanshujha7, this design looks amazing!
However, I think it would be better to keep the same filter design approach used on the https://meshery.io/catalog/models/ page. Having both the approach proposed in this PR and the existing filter design on the Models page could make the UI inconsistent.
I think we should choose one approach. Since @sangramrath suggested using the same approach as on the Models page, I lean toward that option. It would also make the design more familiar and consistent for users.

Himanshujha7:
@Katotodan 
Thanks! This was the suggestion I actually wanted to know.
Sure, I will go ahead as you have suggested and keep it consistent as per @sangramrath's suggestion.

Himanshujha7:
Hi @Katotodan @banana-three-join @Bhumikagarggg 

As per the discussion in Monday's website meeting, I have implemented the suggestions based on the @sangramrath's feedback.

https://github.com/user-attachments/assets/05a93d8e-7718-4202-be32-a9201811601b

-> Reverted the changes from vertical navigator.
-> I have added the category sidebar similar to [https://meshery.io/catalog/models/](https://meshery.io/catalog/models/)
-> Added a filter icon in mobile view for the same since this was not in the source catatlog/models page.
-> Added a pop-up menu containing the category section in mobile view.

Please see the above demo and provide any review if possible.

collections/_pages/extensions.html
```diff
@@ -1,188 +1,473 @@
...
+    function positionPanel() {
+      if (window.innerWidth > STACK_BREAKPOINT) return;
+
+      var margin = 8;
+      var rect = toggle.getBoundingClientRect();
+      var maxLeft = window.innerWidth - panel.offsetWidth - margin;
+
+      panel.style.top = rect.bottom + margin + "px";
+      panel.style.left = Math.max(margin, Math.min(rect.left, maxLeft)) + "px";
+    }
```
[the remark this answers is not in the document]
Himanshujha7: I have addressed these changes in the latest commit.
Please review and let me know

_sass/rootvariables.scss
```diff
@@ -1,118 +1,119 @@
...
-    --action-color-light: #FFF3C5;
-
-    --color-summary-background:#333;
-    --color-summary-background-hover:#666666;
-    --color-details-background:#282828;
-    --color-components-background: #202020;
-}
+:root{
+
+    // Colors that will not change in the dark mode.
```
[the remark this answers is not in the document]
Himanshujha7: Hi, Thanks for the review.
I have added the necessary line space as suggested in latest commit.

collections/_pages/extensions.html
```diff
@@ -1,188 +1,477 @@
...
+          repositionQueued = false;
+        });
+      },
+      { passive: true }
+    );
+  })();
+
+  const extensionCards = document.querySelectorAll(".extensionCard");
+  document.addEventListener("DOMContentLoaded", () => {
+    extensionCards.forEach((card) => {
```
Utkarsh-Mauryaa: @Himanshujha7  `extensionCards.forEach(card => card.style.display = "block")` is redundant here because applyExtensionFilters() already iterates through all cards and sets their display property based on the initial filter/search state. Removing this loop avoids duplicate DOM mutations on page load.

_sass/catalog.scss
```diff
@@ -1,1280 +1,1567 @@
-@use "./variables.scss" as *;
```
Utkarsh-Mauryaa, line 1: @Himanshujha7  Could you please check why there are so many lines of code change in `_sass/catalog.scss` and `_sass/rootvariables.scss`?

It looks like the entire files may have been reformatted, possibly due to editor format-on-save or line-ending changes. Could you please clean this up so the PR only contains the new styles and the changes required for the category sections?

This will also make the PR much easier for reviewers to review and help them focus on the actual functional changes rather than formatting differences.

PARTH-TUSSLE:
Hey @Himanshujha7 ,One thing I'd recommend verifying before approval is the mobile filter panel positioning on shorter viewport heights. Please ensure the panel remains fully within the viewport when opened, rather than extending below the visible area. Other than that, the latest implementation looks much more consistent with the existing Models catalog UI.

collections/_pages/extensions.html
```diff
@@ -1,188 +1,477 @@
...
+{% comment %}
+  Keep only the categories that have something to list, so the navigator never
+  links to an empty section.
+{% endcomment %}
+{% assign categories = "" | split: "" %}
+{% for type in all_types %}
+  {% assign type_count = 0 %}
+  {% for extension in site.extensions %}
+    {% if extension.type == type %}{% if extension.extensionInfo or extension.patternInfo and extension.status !=
+    "coming-soon" %}
```
ishwar170695: I might be missing something here, but I noticed `kubectl-kanvas-snapshot.md` has both `status: coming-soon` and `extensionInfo`. With the current condition, it seems to still be included because `extensionInfo` is truthy, while `status != "coming-soon`" only applies to the `patternInfo` branch. Since the card template and sorting logic also seem to handle `coming-soon` extensions, I just wanted to check if this behavior is intentional.

Himanshujha7: Hey @ishwar170695 
Thanks for the review.
Sorry I couldn't get back to you sooner.

Thanks for pointing it out. I checked, and you are right that status != "coming-soon" only applies to the patternInfo branch.  That said, that clause appears dead either way: no file in `collections/_extensions/` defines `patternInfo` (0 of 25), and all 25 define `extensionInfo ', so I think the condition admits every extension regardless.

Additionally, I would like to point out that It's also pre-existing, not introduced here. 
Master actually had `extension.Status != "ComingSoon"`, which never matched the front matter's `status: coming-soon`, so the clause was already inert. `kubectl-kanvas-snapshot`.

But still, I would be happy to simplify to `{% if extension.extensionInfo or extension.patternInfo %}` so it stops implying a filter that isn't running.

Do let me know your thought to this.

Himanshujha7:
hey @PARTH-TUSSLE 
So i checked the deployed preview and you are right, I have tried to fix this over in my latest commit.
Please let me knwo your thoughts to this.

_sass/catalog.scss
Himanshujha7: Hi @Utkarsh-Mauryaa 
Sorry, I could not get back to you sooner.

I checked on this a few days ago, and thanks for letting me know. The cause was CRLF line endings plus editor reformatting.

I will push with new changes to fix upon this.

collections/_pages/extensions.html
Himanshujha7: Hey @Utkarsh-Mauryaa 

I have removed the duplicate function from here. Will push the updated changes.
Sorry for getting back to this late.

Thanks for pointing it out!

dhruveshmishra (commented):
There are some conflicts please resolve them

collections/_pages/extensions.html
```diff
@@ -1,191 +1,492 @@
...
+        window.requestAnimationFrame(function () {
+          positionPanel();
+          repositionQueued = false;
+        });
+      },
+      { passive: true }
+    );
+  })();
+
+  const extensionCards = document.querySelectorAll(".extensionCard");
```
Utkarsh-Mauryaa: @Himanshujha7 This line should be removed as it is not used anywhere else.

collections/_pages/extensions.html
ishwar170695: Thanks for checking and for the detailed explanation. That clarifies the current behavior. Since the `status` check is pre-existing and `patternInfo` is currently unused, my concern isn't blocking for this PR.

sangramrath:
I suggest we sort A-Z by default on load. This will bring Academy and Collaboration to the top which are extensions we want to show first. 

@Himanshujha7 there's a DCO issue as well that needs fixed.

Himanshujha7:
Thanks for the review, @sangramrath 
Sure, that makes sense - initially I was trying to do something similar in the older design, but yes, I can think of implementing the same here as well.

Yeah, I forgot to sign off on one commit. I'll clean it.

collections/_pages/extensions.html
Himanshujha7: Hey @Utkarsh-Mauryaa 
Yes definitely. This is unused, and I have removed it now over the latest commit.
Thanks!

Himanshujha7:
Hey @sangramrath 
I have made the necessary changes:
Previously -> Sort only applied to each category's entities.
Current -> Sort applies to categories as well as entities in each category.

Have also rebased the branch to correctly sign it off.
Kindly review and let me know if further changes are needed.

collections/_pages/extensions.html
Himanshujha7: Thanks Ishwar

banana-three-join (changes requested):
Two things I'd like to see before merge, plus some smaller notes.

**Line endings on `collections/_pages/extensions.html`**

This file is still CRLF, so the diff shows 685 changed lines when the real change is around 360. Re-diffing with `diff --strip-trailing-cr` confirms there's no reformatting underneath, just the carriage returns.

```bash
sed -i 's/\r$//' collections/_pages/extensions.html
```

**The `coming-soon` guard doesn't filter anything**

Lines 88, 169–170 and 206–207:

```liquid
{% if extension.extensionInfo or extension.patternInfo and extension.status != "coming-soon" %}
```

Liquid has no operator precedence and evaluates right-to-left, so this resolves as `extensionInfo OR (patternInfo AND status != "coming-soon")`. The status check is only reached when `extensionInfo` is absent. `kubectl-kanvas-snapshot.md` is the only extension with `status: coming-soon` and it has `extensionInfo`, so it renders anyway — and now counts toward the Configuration badge that's supposed to exclude it.

If coming-soon should be hidden, force the grouping and compute it once instead of repeating the expression three times:

```liquid
{% assign displayable = false %}
{% if extension.extensionInfo or extension.patternInfo %}
  {% unless extension.status == "coming-soon" %}{% assign displayable = true %}{% endunless %}
{% endif %}
```

If they should be visible, just drop the status clause. That may be the better read given `applySort()` deliberately sorts them last, which only makes sense if they're on the page. Either works — right now the template says it filters them and doesn't.

**Smaller notes**

- Descending sort reverses the section order but the sidebar checkboxes stay alphabetical, so the two disagree after a Z-A sort.
- `scroll-margin-top: 7rem` on `.extension-category` is left over from the reverted navigator; nothing scrolls to those anchors now.
- Extensions without a `type` are dropped by the `{% if extension.type %}` on line 89. Nothing breaks today since all 25 files have one, but an "Other" bucket would match what the comment above it implies.
- Badge counts re-scan `site.extensions` once per category — `site.extensions | group_by: "type"` gets them in one pass.
- The mobile panel handles Escape and outside-click well, but focus never moves into it on open. `role="dialog"` plus focusing the first checkbox would round it out.
- The toolbar sits at 75% of the full container while the results column is offset by the 15rem sidebar, so the search box doesn't line up with the card grid. Worth a side-by-side against `/catalog/models`.

PR #2919, opened 2026-08-28
feat: add extended period filters to leaderboard

shouryapratap132006:
## Proposed changes

Adds a yearly timeframe filter to the Meshery Community Leaderboard and updates period selector labels with dynamic date ranges to match Discourse UI conventions.

### Changes

- Added `yearly` to the leaderboard data generation script (`.github/scripts/fetch-leaderboard.js`).
- Added the `yearly` timeframe option to `_includes/leaderboard.html`.
- Updated period selector options to display dynamic, client-side date ranges:
  - **Week**: `Week <Start Date> – <End Date>` *(e.g. Week Aug 26 – Sep 2)*
  - **Month**: `Month <Start Date> – <End Date>` *(e.g. Month Aug 3 – Sep 2)*
  - **Year**: `Year <Start Date, Year> – <End Date>` *(e.g. Year Sep 2, 2025 – Sep 2)*
  - **All Time**: `All Time`
- Updated `leaderboard.md` introductory description text to reflect `week, month, year, or all-time standings`.
- Preserved existing Discourse period API parameters (`weekly`, `monthly`, `yearly`, `all`) and leaderboard scoring formula (`posts + likes * 2 + solutions * 3`).

### Period semantics

The leaderboard queries Discourse's directory items API using rolling timeframes:

- `weekly` → Last 7 days
- `monthly` → Last 30 days
- `yearly` → Last 365 days
- `all` → All time

### Screen Recording 

https://github.com/user-attachments/assets/426c9262-5305-4d05-9381-e4c1fb20ac39

### Validation

- Tested local Jekyll site preview (`http://127.0.0.1:4000/community/leaderboard/`).
- Verified dynamic date calculations update cleanly based on the client date.
- Ran ESLint (`npx eslint .github/scripts/fetch-leaderboard.js`) — Passed.
- Verified Git diff format & formatting checks (`git diff --check`) — Passed.
- DCO sign-off included.


hiyach28 (changes requested):
please adress the coderabbit comments

shouryapratap132006:
Hi! @hiyach28  I did a deeper investigation into the calendar-year concern raised in the review.

The original implementation used Discourse's:

directory_items.json?period=yearly

However, I verified that on the deployed Discourse version, period=yearly represents a rolling 365-day window (1.year.ago → now), not January 1 → now.

I investigated the available public endpoints, GitHub Actions secrets/references, existing Discourse authentication, Meshery/Discourse integrations, and potential server-side endpoints/plugins. The repository currently has no authenticated Discourse API capability or server-side endpoint that can retrieve posts, likes received, and solutions for January 1 00:00:00 UTC → now using the same time window.

I intentionally did not create an approximation, since that would require mixing calendar-year metrics with rolling-year metrics or estimating dated likes/solutions, which would make the leaderboard inaccurate.

To keep the data authoritative and internally consistent, I updated the UI wording from “This Year” to “Last 12 Months”, which accurately reflects the Discourse period=yearly semantics. All three metrics continue to come from the same directory_items record and therefore use the same rolling 365-day window.

Weekly, monthly, and all-time behavior remain unchanged, and the existing score formula and leaderboard schema are preserved.

A true January 1 → now implementation would require a new capability, such as an authenticated Discourse API mechanism or a custom server-side/Data Explorer endpoint that can aggregate all three metrics over the same calendar-year boundary.

Happy to update the implementation to a true calendar-year period if such a data source can be provided.

hiyach28:
If the metric "This week" and "This month" are used to represrent the period frpm the start of a week (monday) or start of moth (1st), then i think we should follow the same convention for year-wise too.
what do you think @Maanvi212006

Maanvi212006:
Weekly and monthly are also rolling windows (`1.week.ago`, `1.month.ago`), not calendar-aligned (Monday or the 1st). So "Last 12 Months" is the correct label here. Unless we're explicitly asked to switch to calendar-aligned boundaries (Monday/1st/Jan 1) across the board, we should stay consistent with the convention already used for weekly and monthly, and apply the same for "This Year."

shouryapratap132006:
Hi @Maanvi212006 @hiyach28 , I’ve addressed the review feedback and kept the yearly leaderboard consistent with the existing rolling-period behaviour. The UI now clearly labels it as “Last 12 Months”, and the introductory text refers to the selected period.

All checks are passing, including DCO. Could you please take a final look and approve/merge when convenient? Thanks! 🙌

.github/scripts/fetch-leaderboard.js
banana-three-join, line 1: The added comments are not needed. The functions are already self-explanatory so please revert on that. Also, regarding the `Use a calendar-year boundary for yearly` concern, I believe it's invalid because the fetch is being done at `${BASE_URL}?period=${period}&order=likes_received&limit=50` which leads me to believe that `discuss.meshery.io` already handles this through set periods and not from when it's requested to 7, 30, 365 days back. If that isn't the case then the way on how the periods are being fetched would need to be updated since this wouldn't be a particular behavior to just the `yearly` fetches.

_includes/leaderboard.html
```diff
@@ -5,6 +5,7 @@
             <select id="period-select" class="period-select" aria-label="Leaderboard time period">
                 <option value="weekly">This Week</option>
                 <option value="monthly">This Month</option>
+                <option value="yearly">Last 12 Months</option>
```
banana-three-join: `Last 12 months` isn't the same as yearly. It should be Current year. 

Overall, I think the wording to `Current` instead of this `This` might be better. Thoughts? @Maanvi212006

banana-three-join (changes requested):

banana-three-join:
Just saw the conversation above after I brought up the same concern in my review of the PR. If all of them are rolling windows then an issue must be started to update the fetching mechanism, if one of you could please do so. @hiyach28 @Maanvi212006 

Really appreciate it!

Maanvi212006:
> Just saw the conversation above after I brought up the same concern in my review of the PR. If all of them are rolling windows then an issue must be started to update the fetching mechanism, if one of you could please do so. @hiyach28 @Maanvi212006
> 
> Really appreciate it!

Alright @banana-three-join , I will verify and open the issue!

shouryapratap132006:
Hi @Maanvi212006, the latest changes address the review feedback and keep the yearly leaderboard consistent with Discourse's existing rolling-period behaviour. The UI now uses “Last 12 Months”, and the introductory text refers to the selected period.

All checks are passing and there are no remaining actionable review comments. When you get a chance, could you please take a final look and approve/merge? Thanks! 🙌

Maanvi212006:
I confirmed via Discourse's source (`directory_item.rb`) that `weekly`/`monthly`/`yearly` are rolling windows (`1.week.ago`, etc.), not calendar-aligned. And the API only returns pre-aggregated totals, no raw timestamps to re-bucket ourselves. 

One way, to get calendar-aligned data could be to fetch raw, timestamped activity per user (not the pre-aggregated `directory_items.json` totals), then compute calendar-aligned counts ourselves, eg. using an endpoint that returns individual likes/posts/solutions with dates, filtering by `created_at >= startOfWeek/Month/Year`, and summing manually. But, this likely means one API call per user instead of the current single call per period. But, it would result into hundreds of requests for an active forum vs the 3 we make today. Plus, there is a risk of hitting Discourse's rate limits. 
There is one more option but that would require admin API access. 

As per me, for this PR, we can stick to "12 months from now", also we can expand the scope of the PR to rename other labels to "Last 7 days", "Last 30 days". And then, we can discuss about the calendar-aligned approach. 

Thoughts @banana-three-join @hiyach28 ? Or, any other approach we can think of?

Maanvi212006:
It has been decided to use "last 7/30/365 days" label itself, as displaying calendar-aligned data is beyond the scope of our fix.

So, @shouryapratap132006 we will be expanding the scope of this PR. You need to rename the labels of "This month" and "This week" to "Last 7 days" and "Last 30 days" respectively. Also, use "Last 356 days" instead of the "Yearly" label. 

Update the PR title and description and push the necessary changes. I will update the issue details as well.

shouryapratap132006:
Hi @Maanvi212006, thanks for the clarification. I've updated the period labels to reflect the rolling windows: Last 7 Days, Last 30 Days, and Last 365 Days. I've also updated the PR title/description accordingly and removed the unnecessary comments as requested.

The implementation continues to use the existing Discourse period parameters, with calendar-aligned periods remaining out of scope. Please take a look when convenient. Thanks! 🙌

Maanvi212006 (commented):
Please attach a screen recording of the changes you made. Also, I have updated the issue description, please go through it again

shouryapratap132006:
Hi @Maanvi212006 , I have added the screen recording in PR description and made changes according to the new updated issue, Please take a look when convenient. Thanks!

Maanvi212006 (commented):
Looks good! But I believe the dropdown opening below the box (and not covering it) would look better (even though the design I shared has the one like you did, i.e. covering). Thoughts @banana-three-join @hiyach28 ?

image

banana-three-join:
@Maanvi212006 I agree. Also, there needs to be a visual text separation between time "period" and "range"

banana-three-join:
Also, the `leaderboard.json` can be omitted from the merge

shouryapratap132006:
Hi @Maanvi212006 @hiyach28 @banana-three-join , I’ve addressed the latest review feedback on the leaderboard filter.

Updated the period dropdown to open below the selector without covering the selected period.
Added clear visual separation between the period name and its date range.
Kept the existing rolling-period semantics (weekly, monthly, yearly, all).
Removed _data/leaderboard.json from the PR changes as suggested.
Preserved the existing leaderboard functionality and dark-mode styling.

I also validated the changes with a successful Jekyll build and ESLint check, and verified that only the intended leaderboard files are modified.

Could you please take another look when convenient? Thanks! 🙌

Maanvi212006 (changes requested):
Looks good, just few things:

1. `_data/leaderboard.json` is still in this PR

2. Yearly data has 13 tie-break ordering violations : `buildLeaderboard()`'s sort is correct. But the committed `periods.yearly` array in `_data/leaderboard.json` doesn't reflect it. The data needs to be regenerated by the updated script

3. No need to make changes to `leaderboard.md` (its already covered in #2911)

Also, attach the screen recording of the latest changes in the description

shouryapratap132006:
Hi @Maanvi212006, I’ve addressed the latest review feedback:

Removed _data/leaderboard.json from the PR.
Kept the yearly period generation in fetch-leaderboard.js.
Verified the regenerated yearly data has 0 tie-break violations.
Confirmed leaderboard.md remains unchanged as requested.
Preserved the dropdown positioning and period/date-range UI improvements.
All validation checks are passing, including ESLint, Jekyll build, and diff checks.

Could you please take another look when convenient? Thanks! 🙌

Maanvi212006 (commented):
Yes, the PR looks good now! Can you update the screen recording in the PR description to reflect the updated changes though?

shouryapratap132006:
Hi @Maanvi212006 , added the screen recording please check and verify.

Maanvi212006 (dismissed):
LGTM!

banana-three-join (commented):
Thanks for the work on this! I do have a comment regarding the current approach.

**The custom select**

The new dropdown isn't needed for the feature itself. It's needed for one styling choice — rendering the period name and the date range as two separately styled spans — and a native `<option>` can't hold that. If the label is flattened into a single string like `Week · Aug 26 – Sep 2`, the existing `<select>` can stay, and `appearance: none` plus a background-image chevron gives full control over how the closed control looks in every browser. The dropdown panel stays browser-drawn, which seems like a fair trade here.

What makes me want to push on this is that the new component doesn't really do anything on its own — it writes to the old select and dispatches a `change` event so the existing code keeps working. That leaves two controls in the DOM for one input, and it means reimplementing arrow keys, typeahead, and focus handling that the native element already had.

I don't think the two-part label is a firm requirement, so. Is there anything else which warrants the creation of a custom element? If not, then it can be removed the stylization can fall back onto the current `select` element.

**Repeated classes**

A lot of the classes are repeating across elements in a way that makes the markup hard to read, and I think most of them can come out with better CSS rules rather than more hooks:

- Each wrapper has an `id` identical to its `class` (`custom-period-select`, `period-select-trigger`, `period-select-dropdown`). One or the other is enough.
- `period-option-name` and `period-option-range` repeat on every option to mean "first child" and "second child". `.period-option > :first-child` and `.period-option [data-range]` cover both, and `data-range` is already on the element.
- State is stored twice — `selected` alongside `aria-selected`, and `open` alongside `aria-expanded`. Styling off the ARIA attributes would let the class toggling come out of the JS entirely.

These two points are related: if the first one lands, most of the second goes away on its own.

banana-three-join (approved):
LGTM. Ty to the both of you, @Maanvi212006 @shouryapratap132006!

PR #2923, opened 2026-08-29
feat: generate GitHub contributor activity statistics for community leaderboard

### LINKED ISSUE #2917: [feat] Generate GitHub contributor activity statistics for the Community Leaderboard
### Current Behavior

The [Meshery Community Leaderboard](https://meshery.io/community/leaderboard/) currently surfaces community activity from the Meshery discussion forum.
The leaderboard already uses a scheduled GitHub Actions workflow to generate and persist leaderboard data within `meshery/meshery.io`. The existing implementation can be referenced in [PR #2767](https://github.com/meshery/meshery.io/pull/2767) 
GitHub activity is currently not collected as part of this data. Contributions such as opening issues, opening pull requests, and participating in GitHub discussions are therefore not represented in the leaderboard data.

### Expected Behavior

Extend the existing leaderboard data-generation workflow to collect GitHub contributor activity and render it as a separate leaderboard board on the same Community Leaderboard page; alongside, not merged with, the existing discussion-forum board. The two boards track different stats and produce independent rankings; no mapping or tying of data between sources.
The following GitHub activity should be captured (at minimum):

- Issues opened
- Pull requests opened
- Pull requests reviewed (comments made)

The generated data should be stored in a machine-readable format that can be consumed by the Community Leaderboard.

---
### Contributor [Guides](https://docs.meshery.io/project/contributing)

yagyaXsen:
## Description

This PR addresses issue [#2917](https://github.com/meshery/meshery.io/issues/2917) by extending the Community Leaderboard with GitHub contributor activity statistics.

### GitHub contributor activity

The Community Leaderboard now includes a separate GitHub activity board alongside the existing Discourse leaderboard on the same page.

The two boards are independent:

- Different activity metrics
- Independent rankings
- No mapping or tying of contributor data between GitHub and Discourse

The existing Discourse leaderboard score and ranking remain unchanged.

### GitHub metrics

The GitHub leaderboard tracks:

- **Issues opened**
- **Pull requests merged**
- **Pull requests reviewed**

GitHub contributors are ranked using:

GitHub Score = (3 × PRs merged) + (2 × Issues opened) + (1 × PRs reviewed)

## Changes

- **`.github/scripts/fetch-leaderboard.js`**: Extended to fetch GitHub activity for each contributor via the GitHub REST API. Uses exact review timestamps for accurate period filtering. Rate-limit-safe delays included.
- **`.github/workflows/update-leaderboard.yml`**: Passes `GITHUB_TOKEN` (read-only) to the fetch script via env.
- **`_includes/leaderboard.html`**: Added a separate GitHub activity board (table and podium cards) alongside the existing Discourse board. 
- **`_sass/leaderboard.scss`**: Added styles for new GitHub stat sections utilizing existing CSS variables.
- **`_data/github_leaderboard.json`**: A generated JSON snapshot of GitHub contributor stats 

https://github.com/user-attachments/assets/fa4335c0-3d7e-4d15-95ad-c7792be0aeec

## How to test locally

```bash
export GITHUB_TOKEN=ghp_your_token_here
node .github/scripts/fetch-leaderboard.js
cat _data/leaderboard.json | jq '.periods.all[0]'
```

## Checklist
- [x] Code follows the existing style
- [x] No new npm dependencies added
- [x] No trailing whitespace (`git diff --check` clean)
- [x] Existing Discourse leaderboard data is untouched
- [x] GitHub API calls have per-user error handling
- [x] Rate limit safe delays added between API calls
- [x] Commits signed off with `-s`

Fixes #2917

### DIFF
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -15,6 +15,12 @@ const PERIOD_SINCE = {
   all: null
 };
 
+// Explicit mapping of verified Discourse username to GitHub username.
+// Users not in this list will not have GitHub stats pulled.
+const GITHUB_USERNAME_MAPPING = {
+  "theBeginner86": "theBeginner86"
+};
+
 async function fetchUsers(period) {
   const headers = {
     'User-Agent': 'meshery-leaderboard-bot/1.0',
@@ -110,21 +116,22 @@ async function githubFetch(url, attempt) {
 
 async function fetchPeriodStats(username, since, githubUser) {
   const sinceFilter = since ? `+created:>=${since}` : '';
+  const ghUsername = githubUser.login;
 
   const issuesData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+type:issue${sinceFilter}&per_page=1`
+    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(ghUsername)}+org:${GITHUB_ORG}+type:issue${sinceFilter}&per_page=1`
   );
   const issues = issuesData ? (issuesData.total_count || 0) : 0;
   await sleep(120);
 
   const prsData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilter}&per_page=1`
+    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(ghUsername)}+org:${GITHUB_ORG}+is:pr${sinceFilter}&per_page=1`
   );
   const prs = prsData ? (prsData.total_count || 0) : 0;
   await sleep(120);
 
   const reviewsData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=reviewed-by:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilter}&per_page=1`
+    `${GITHUB_API}/search/issues?q=reviewed-by:${encodeURIComponent(ghUsername)}+org:${GITHUB_ORG}+is:pr${sinceFilter}&per_page=1`
   );
   const reviews = reviewsData ? (reviewsData.total_count || 0) : 0;
 
@@ -162,8 +169,13 @@ async function buildAllPeriods() {
 
   const userCache = {}; // keyed by username -> { login, html_url } or null
   for (const username of allUsernames) {
+    const mappedGithubUser = GITHUB_USERNAME_MAPPING[username];
+    if (!mappedGithubUser) {
+      userCache[username] = null;
+      continue;
+    }
     try {
-      const user = await githubFetch(`${GITHUB_API}/users/${encodeURIComponent(username)}`);
+      const user = await githubFetch(`${GITHUB_API}/users/${encodeURIComponent(mappedGithubUser)}`);
       userCache[username] = user;
       await sleep(100);
     } catch (err) {
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -3,6 +3,59 @@ const path = require('path');
 
 const BASE_URL = 'https://discuss.meshery.io/directory_items.json';
 const PERIODS = ['weekly', 'monthly', 'all'];
+const GITHUB_API = 'https://api.github.com';
+const GITHUB_ORG = 'meshery';
+const GITHUB_TOKEN = process.env.GITHUB_TOKEN || process.env.GH_ACCESS_TOKEN || '';
+
+function sleep(ms) {
+  return new Promise(r => setTimeout(r, ms));
+}
+
+async function githubFetch(url) {
+  const headers = {
+    'User-Agent': 'meshery-leaderboard-bot/1.0',
+    'Accept': 'application/vnd.github+json',
+    'X-GitHub-Api-Version': '2022-11-28',
+  };
+  if (GITHUB_TOKEN) headers['Authorization'] = `Bearer ${GITHUB_TOKEN}`;
+  const res = await fetch(url, { headers });
+  if (res.status === 404) return null;
+  if (!res.ok) throw new Error(`GitHub API error: ${res.status} ${url}`);
+  return res.json();
+}
+
+async function fetchGitHubStats(username) {
+  const user = await githubFetch(`${GITHUB_API}/users/${encodeURIComponent(username)}`);
+  if (!user) return null;
+
+  await sleep(120);
+
+  const issuesData = await githubFetch(
+    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+type:issue&per_page=1`
+  );
+  const issues = issuesData ? (issuesData.total_count || 0) : 0;
+  await sleep(120);
+
+  const prsData = await githubFetch(
+    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr&per_page=1`
+  );
+  const prs = prsData ? (prsData.total_count || 0) : 0;
+  await sleep(120);
+
+  const reviewsData = await githubFetch(
+    `${GITHUB_API}/search/issues?q=reviewed-by:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr&per_page=1`
+  );
+  const reviews = reviewsData ? (reviewsData.total_count || 0) : 0;
+
+  return {
+    github_username: user.login,
+    github_profile_url: user.html_url,
+    github_issues: issues,
+    github_prs: prs,
+    github_reviews: reviews,
+    github_score: (issues * 2) + (prs * 3) + (reviews * 2),
+  };
+}
 
 async function fetchUsers(period) {
   const headers = {
@@ -61,6 +114,43 @@ async function buildAllPeriods() {
 
   if (!periods.all.length) {
     throw new Error('all-time period empty — refusing to overwrite');
+  }
+
+  const allUsernames = [...new Set(periods.all.map(u => u.username))];
+  const githubCache = {};
+
+  for (const username of allUsernames) {
+    try {
+      const stats = await fetchGitHubStats(username);
+      githubCache[username] = stats;
+      console.log(`GitHub stats for ${username}:`, stats ? `issues=${stats.github_issues} prs=${stats.github_prs} reviews=${stats.github_reviews}` : 'not found');
+    } catch (err) {
+      console.warn(`Failed GitHub stats for ${username}: ${err.message}`);
+      githubCache[username] = null;
+    }
+    await sleep(200);
+  }
+
+  for (const period of PERIODS) {
+    periods[period] = periods[period].map(user => {
+      const gh = githubCache[user.username];
+      const github_issues = gh ? Math.max(0, parseInt(gh.github_issues) || 0) : 0;
+      const github_prs = gh ? Math.max(0, parseInt(gh.github_prs) || 0) : 0;
+      const github_reviews = gh ? Math.max(0, parseInt(gh.github_reviews) || 0) : 0;
+      const github_score = gh ? Math.max(0, parseInt(gh.github_score) || 0) : 0;
+      const total_score = user.score + github_score;
+      return {
+        ...user,
+        github_username: gh ? gh.github_username : '',
+        github_profile_url: gh ? gh.github_profile_url : '',
+        github_issues,
+        github_prs,
+        github_reviews,
+        github_score,
+        total_score,
+      };
+    }).sort((a, b) => b.total_score - a.total_score)
+      .map((u, i) => ({ ...u, rank: i + 1 }));
   }
 
   return periods;
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -10,9 +10,9 @@ const GITHUB_API = 'https://api.github.com';
 const GITHUB_ORG = 'meshery';
 const GITHUB_TOKEN = process.env.GITHUB_TOKEN || process.env.GH_ACCESS_TOKEN || '';
 const PERIOD_SINCE = {
-  weekly:  new Date(Date.now() - 7  * 24 * 60 * 60 * 1000).toISOString().slice(0, 10),
+  weekly: new Date(Date.now() - 7 * 24 * 60 * 60 * 1000).toISOString().slice(0, 10),
   monthly: new Date(Date.now() - 30 * 24 * 60 * 60 * 1000).toISOString().slice(0, 10),
-  all:     null
+  all: null
 };
 
 async function fetchUsers(period) {
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -108,42 +108,29 @@ async function githubFetch(url, attempt) {
   return res.json();
 }
 
-async function fetchGitHubStats(username, since) {
-  const user = await githubFetch(`${GITHUB_API}/users/${encodeURIComponent(username)}`);
-  if (!user) return null;
-
-  await sleep(120);
-
-  const sinceFilterForIssues = since ? `+created:>=${since}` : '';
-  const sinceFilterForReviews = since ? `+merged:>=${since}` : '';
+async function fetchPeriodStats(username, since, githubUser) {
+  const sinceFilter = since ? `+created:>=${since}` : '';
 
   const issuesData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+type:issue${sinceFilterForIssues}&per_page=1`
+    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+type:issue${sinceFilter}&per_page=1`
   );
   const issues = issuesData ? (issuesData.total_count || 0) : 0;
   await sleep(120);
 
   const prsData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilterForIssues}&per_page=1`
+    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilter}&per_page=1`
   );
   const prs = prsData ? (prsData.total_count || 0) : 0;
   await sleep(120);
 
   const reviewsData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=reviewed-by:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilterForReviews}&per_page=1`
+    `${GITHUB_API}/search/issues?q=reviewed-by:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilter}&per_page=1`
   );
-  const prsReviewedCount = reviewsData ? (reviewsData.total_count || 0) : 0;
-  await sleep(120);
-
-  const commentReviewsData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=commenter:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilterForReviews}&per_page=1`
-  );
-  const reviewCommentsCount = commentReviewsData ? (commentReviewsData.total_count || 0) : 0;
-  const reviews = prsReviewedCount + reviewCommentsCount;
+  const reviews = reviewsData ? (reviewsData.total_count || 0) : 0;
 
   return {
-    github_username: user.login,
-    github_profile_url: user.html_url,
+    github_username: githubUser.login || '',
+    github_profile_url: githubUser.html_url || '',
     github_issues: issues,
     github_prs: prs,
     github_reviews: reviews,
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -1,3 +1,6 @@
+/* eslint-env node */
+'use strict';
+
 const fs = require('fs');
 const path = require('path');
 
@@ -6,56 +9,11 @@ const PERIODS = ['weekly', 'monthly', 'all'];
 const GITHUB_API = 'https://api.github.com';
 const GITHUB_ORG = 'meshery';
 const GITHUB_TOKEN = process.env.GITHUB_TOKEN || process.env.GH_ACCESS_TOKEN || '';
-
-function sleep(ms) {
-  return new Promise(r => setTimeout(r, ms));
-}
-
-async function githubFetch(url) {
-  const headers = {
-    'User-Agent': 'meshery-leaderboard-bot/1.0',
-    'Accept': 'application/vnd.github+json',
-    'X-GitHub-Api-Version': '2022-11-28',
-  };
-  if (GITHUB_TOKEN) headers['Authorization'] = `Bearer ${GITHUB_TOKEN}`;
-  const res = await fetch(url, { headers });
-  if (res.status === 404) return null;
-  if (!res.ok) throw new Error(`GitHub API error: ${res.status} ${url}`);
-  return res.json();
-}
-
-async function fetchGitHubStats(username) {
-  const user = await githubFetch(`${GITHUB_API}/users/${encodeURIComponent(username)}`);
-  if (!user) return null;
-
-  await sleep(120);
-
-  const issuesData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+type:issue&per_page=1`
-  );
-  const issues = issuesData ? (issuesData.total_count || 0) : 0;
-  await sleep(120);
-
-  const prsData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr&per_page=1`
-  );
-  const prs = prsData ? (prsData.total_count || 0) : 0;
-  await sleep(120);
-
-  const reviewsData = await githubFetch(
-    `${GITHUB_API}/search/issues?q=reviewed-by:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr&per_page=1`
-  );
-  const reviews = reviewsData ? (reviewsData.total_count || 0) : 0;
-
-  return {
-    github_username: user.login,
-    github_profile_url: user.html_url,
-    github_issues: issues,
-    github_prs: prs,
-    github_reviews: reviews,
-    github_score: (issues * 2) + (prs * 3) + (reviews * 2),
-  };
-}
+const PERIOD_SINCE = {
+  weekly:  new Date(Date.now() - 7  * 24 * 60 * 60 * 1000).toISOString().slice(0, 10),
+  monthly: new Date(Date.now() - 30 * 24 * 60 * 60 * 1000).toISOString().slice(0, 10),
+  all:     null
+};
 
 async function fetchUsers(period) {
   const headers = {
@@ -87,17 +45,16 @@ function buildLeaderboard(items) {
     .map(item => ({
       rank: 0,
       username: item.user.username,
-      avatar: item.user.avatar_template ? `https://discuss.meshery.io${item.user.avatar_template.replace('{size}', '200')}` : '',
+      avatar: item.user.avatar_template
+        ? `https://discuss.meshery.io${item.user.avatar_template.replace('{size}', '200')}`
+        : '',
       profile_url: `https://discuss.meshery.io/u/${item.user.username}`,
       posts: item.post_count || 0,
       likes: item.likes_received || 0,
       solutions: item.solutions || 0,
       score: computeScore(item)
     }))
-    .sort((a, b) => {
-      if (b.score !== a.score) return b.score - a.score;
-      return a.username.localeCompare(b.username);
-    })
+    .sort((a, b) => b.score - a.score)
     .map((user, index) => ({
       ...user,
       rank: index + 1
@@ -116,40 +73,60 @@ async function buildAllPeriods() {
     throw new Error('all-time period empty — refusing to overwrite');
   }
 
-  const allUsernames = [...new Set(periods.all.map(u => u.username))];
+  const allUsernames = [
+    ...new Set([
+      ...periods.weekly.map(u => u.username),
+      ...periods.monthly.map(u => u.username),
+      ...periods.all.map(u => u.username)
+    ])
+  ];
   const githubCache = {};
 
-  for (const username of allUsernames) {
-    try {
-      const stats = await fetchGitHubStats(username);
-      githubCache[username] = stats;
-      console.log(`GitHub stats for ${username}:`, stats ? `issues=${stats.github_issues} prs=${stats.github_prs} reviews=${stats.github_reviews}` : 'not found');
-    } catch (err) {
-      console.warn(`Failed GitHub stats for ${username}: ${err.message}`);
-      githubCache[username] = null;
+  console.log(`Fetching GitHub stats for ${allUsernames.length} users...`);
+
+  for (const period of PERIODS) {
+    const since = PERIOD_SINCE[period];
+    for (const username of allUsernames) {
+      const cacheKey = `${username}:${period}`;
+      if (githubCache[cacheKey] !== undefined) continue;
+      try {
+        const stats = await fetchGitHubStats(username, since);
+        githubCache[cacheKey] = stats;
+        console.log(
+          `GitHub stats for ${username} (${period}):`,
+          stats
+            ? `issues=${stats.github_issues} prs=${stats.github_prs} reviews=${stats.github_reviews}`
+            : 'not found on GitHub'
+        );
+      } catch (err) {
+        console.warn(`Failed GitHub stats for ${username} (${period}): ${err.message}`);
+        githubCache[cacheKey] = null;
+      }
+      await sleep(200);
     }
-    await sleep(200);
   }
 
   for (const period of PERIODS) {
-    periods[period] = periods[period].map(user => {
-      const gh = githubCache[user.username];
-      const github_issues = gh ? Math.max(0, parseInt(gh.github_issues) || 0) : 0;
-      const github_prs = gh ? Math.max(0, parseInt(gh.github_prs) || 0) : 0;
-      const github_reviews = gh ? Math.max(0, parseInt(gh.github_reviews) || 0) : 0;
-      const github_score = gh ? Math.max(0, parseInt(gh.github_score) || 0) : 0;
-      const total_score = user.score + github_score;
-      return {
-        ...user,
-        github_username: gh ? gh.github_username : '',
-        github_profile_url: gh ? gh.github_profile_url : '',
-        github_issues,
-        github_prs,
-        github_reviews,
-        github_score,
-        total_score,
-      };
-    }).sort((a, b) => b.total_score - a.total_score)
+    periods[period] = periods[period]
+      .map(user => {
+        const gh = githubCache[`${user.username}:${period}`];
+        const github_issues = gh ? (gh.github_issues || 0) : 0;
+        const github_prs = gh ? (gh.github_prs || 0) : 0;
+        const github_reviews = gh ? (gh.github_reviews || 0) : 0;
+        const github_score = gh ? (gh.github_score || 0) : 0;
+        const total_score = (user.score || 0) + github_score;
+        return {
+          ...user,
+          github_username: gh ? (gh.github_username || '') : '',
+          github_profile_url: gh ? (gh.github_profile_url || '') : '',
+          github_issues,
+          github_prs,
+          github_reviews,
+          github_score,
+          total_score
+        };
+      })
+      .sort((a, b) => b.total_score - a.total_score)
       .map((u, i) => ({ ...u, rank: i + 1 }));
   }
 
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -85,7 +85,8 @@ async function githubFetch(url, attempt) {
   // Handle rate limiting
   const isRateLimited =
     res.status === 429 ||
-    (res.status === 403 && res.headers.get('x-ratelimit-remaining') === '0');
+    (res.status === 403 && res.headers.get('x-ratelimit-remaining') === '0') ||
+    (res.status === 403 && res.headers.get('retry-after') !== null);
 
   if (isRateLimited && attempt < 2) {
     const retryAfter = res.headers.get('retry-after');
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -54,11 +54,100 @@ function buildLeaderboard(items) {
       solutions: item.solutions || 0,
       score: computeScore(item)
     }))
-    .sort((a, b) => b.score - a.score)
+    .sort((a, b) => {
+      if (b.score !== a.score) return b.score - a.score;
+      return a.username.localeCompare(b.username);
+    })
     .map((user, index) => ({
       ...user,
       rank: index + 1
     }));
+}
+
+function sleep(ms) {
+  return new Promise(r => setTimeout(r, ms));
+}
+
+async function githubFetch(url, attempt) {
+  if (attempt === undefined) attempt = 0;
+  const headers = {
+    'User-Agent': 'meshery-leaderboard-bot/1.0',
+    'Accept': 'application/vnd.github+json',
+    'X-GitHub-Api-Version': '2022-11-28'
+  };
+  if (GITHUB_TOKEN) {
+    headers['Authorization'] = `Bearer ${GITHUB_TOKEN}`;
+  }
+  const res = await fetch(url, { headers });
+
+  if (res.status === 404) return null;
+
+  // Handle rate limiting
+  const isRateLimited =
+    res.status === 429 ||
+    (res.status === 403 && res.headers.get('x-ratelimit-remaining') === '0');
+
+  if (isRateLimited && attempt < 2) {
+    const retryAfter = res.headers.get('retry-after');
+    const resetAt = res.headers.get('x-ratelimit-reset');
+    let waitMs;
+    if (retryAfter) {
+      waitMs = (parseInt(retryAfter, 10) + 2) * 1000;
+    } else if (resetAt) {
+      waitMs = Math.max(0, (parseInt(resetAt, 10) * 1000) - Date.now()) + 2000;
+    } else {
+      waitMs = 60000; // default: wait 60 seconds
+    }
+    console.warn(`Rate limited by GitHub. Waiting ${Math.round(waitMs / 1000)}s before retry...`);
+    await sleep(waitMs);
+    return githubFetch(url, attempt + 1);
+  }
+
+  if (!res.ok) throw new Error(`GitHub API error: ${res.status} ${url}`);
+  return res.json();
+}
+
+async function fetchGitHubStats(username, since) {
+  const user = await githubFetch(`${GITHUB_API}/users/${encodeURIComponent(username)}`);
+  if (!user) return null;
+
+  await sleep(120);
+
+  const sinceFilterForIssues = since ? `+created:>=${since}` : '';
+  const sinceFilterForReviews = since ? `+merged:>=${since}` : '';
+
+  const issuesData = await githubFetch(
+    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+type:issue${sinceFilterForIssues}&per_page=1`
+  );
+  const issues = issuesData ? (issuesData.total_count || 0) : 0;
+  await sleep(120);
+
+  const prsData = await githubFetch(
+    `${GITHUB_API}/search/issues?q=author:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilterForIssues}&per_page=1`
+  );
+  const prs = prsData ? (prsData.total_count || 0) : 0;
+  await sleep(120);
+
+  const reviewsData = await githubFetch(
+    `${GITHUB_API}/search/issues?q=reviewed-by:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilterForReviews}&per_page=1`
+  );
+  const prsReviewedCount = reviewsData ? (reviewsData.total_count || 0) : 0;
+  await sleep(120);
+
+  const commentReviewsData = await githubFetch(
+    `${GITHUB_API}/search/issues?q=commenter:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilterForReviews}&per_page=1`
+  );
+  const reviewCommentsCount = commentReviewsData ? (commentReviewsData.total_count || 0) : 0;
+  const reviews = prsReviewedCount + reviewCommentsCount;
+
+  return {
+    github_username: user.login,
+    github_profile_url: user.html_url,
+    github_issues: issues,
+    github_prs: prs,
+    github_reviews: reviews,
+    github_score: (issues * 2) + (prs * 3) + (reviews * 2)
+  };
 }
 
 async function buildAllPeriods() {
@@ -80,17 +169,32 @@ async function buildAllPeriods() {
       ...periods.all.map(u => u.username)
     ])
   ];
-  const githubCache = {};
 
   console.log(`Fetching GitHub stats for ${allUsernames.length} users...`);
 
+  const userCache = {}; // keyed by username -> { login, html_url } or null
+  for (const username of allUsernames) {
+    try {
+      const user = await githubFetch(`${GITHUB_API}/users/${encodeURIComponent(username)}`);
+      userCache[username] = user;
+      await sleep(100);
+    } catch (err) {
+      console.warn(`GitHub user lookup failed for ${username}: ${err.message}`);
+      userCache[username] = null;
+    }
+  }
+
+  const githubCache = {}; // keyed by "username:period"
   for (const period of PERIODS) {
     const since = PERIOD_SINCE[period];
     for (const username of allUsernames) {
+      if (!userCache[username]) {
+        githubCache[`${username}:${period}`] = null;
+        continue;
+      }
       const cacheKey = `${username}:${period}`;
-      if (githubCache[cacheKey] !== undefined) continue;
       try {
-        const stats = await fetchGitHubStats(username, since);
+        const stats = await fetchPeriodStats(username, since, userCache[username]);
         githubCache[cacheKey] = stats;
         console.log(
           `GitHub stats for ${username} (${period}):`,
@@ -99,7 +203,7 @@ async function buildAllPeriods() {
             : 'not found on GitHub'
         );
       } catch (err) {
-        console.warn(`Failed GitHub stats for ${username} (${period}): ${err.message}`);
+        console.warn(`Failed period stats for ${username} (${period}): ${err.message}`);
         githubCache[cacheKey] = null;
       }
       await sleep(200);
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -1,6 +1,3 @@
-/* eslint-env node */
-'use strict';
-
 const fs = require('fs');
 const path = require('path');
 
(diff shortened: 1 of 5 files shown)


yagyaXsen:
@coderabbitai Fixed all 4 issues:
1. Added Retry-After / x-ratelimit-reset header handling with retry in githubFetch
2. GitHub stats now scoped to period using created:>= date filter on Search API
3. Review counting now uses both reviewed-by: and commenter: for better accuracy
4. allUsernames now collects from all periods (weekly + monthly + all) not just all-time

yagyaXsen:
Fixed in latest commit — removed extra key spacing in `PERIOD_SINCE` object to satisfy ESLint `key-spacing` rule.

.github/scripts/fetch-leaderboard.js
```diff
@@ -34,18 +45,103 @@ function buildLeaderboard(items) {
...
+    `${GITHUB_API}/search/issues?q=reviewed-by:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilterForReviews}&per_page=1`
+  );
+  const prsReviewedCount = reviewsData ? (reviewsData.total_count || 0) : 0;
+  await sleep(120);
+
+  const commentReviewsData = await githubFetch(
+    `${GITHUB_API}/search/issues?q=commenter:${encodeURIComponent(username)}+org:${GITHUB_ORG}+is:pr${sinceFilterForReviews}&per_page=1`
+  );
+  const reviewCommentsCount = commentReviewsData ? (commentReviewsData.total_count || 0) : 0;
+  const reviews = prsReviewedCount + reviewCommentsCount;
```
hiyach28: please avoid adding `prsReviewedCount` and `reviewCommentsCount` directly here, these two queries overlap: a formal GitHub review also produces review/comment activity, so the same review can be counted in both sets and inflate the `reviews` metric.
could we use the formal `reviewed-by:` count alone here, or otherwise deduplicate the two result sets before calculating this value?

.github/scripts/fetch-leaderboard.js
```diff
@@ -34,18 +45,103 @@ function buildLeaderboard(items) {
...
+}
+
+async function fetchGitHubStats(username, since) {
+  const user = await githubFetch(`${GITHUB_API}/users/${encodeURIComponent(username)}`);
+  if (!user) return null;
+
+  await sleep(120);
+
+  const sinceFilterForIssues = since ? `+created:>=${since}` : '';
+  const sinceFilterForReviews = since ? `+merged:>=${since}` : '';
```
hiyach28: I think this should use a review-appropriate date filter rather than `merged:`. The `merged:` qualifier filters based on when the PR was merged, not when the contributor actually performed the review.

so, a review submitted before the selected period can be counted if the reviewed PR was merged during the period. Could we revisit the period filtering here so the review count is scoped to review activity rather than PR merge activity?

.github/scripts/fetch-leaderboard.js
```diff
@@ -34,18 +45,103 @@ function buildLeaderboard(items) {
...
+  const res = await fetch(url, { headers });
+
+  if (res.status === 404) return null;
+
+  // Handle rate limiting
+  const isRateLimited =
+    res.status === 429 ||
+    (res.status === 403 && res.headers.get('x-ratelimit-remaining') === '0');
+
+  if (isRateLimited && attempt < 2) {
```
hiyach28, line 96: could we make the exhausted rate-limit case fail explicitly rather than allowing this to fall through as a missing/zero-stat result?

hiyach28 (changes requested):
could we consider reducing the repeated github api calls here? current flow appears to fetch the same user's github profile/stat-related data separately for each reporting period, which can multiply the number of requests significantly as the contributor count grows.
atleast, user-level data that doesn't depend on the reporting period could be cached once per username. this would help keep the workflow from approaching limits as the leaderboard scales.

yagyaXsen:
Hi @hiyach28, thank you for the detailed review! All 3 points are valid:

1. **Repeated API calls** — will cache the GitHub user existence check once per 
   username and only run period-scoped queries per period, reducing calls 
   from `users × 3 periods × 4 calls` to `users × (1 user check + 3 periods × 3 calls)`

2. **Double counting reviews** — will drop `commenter:` entirely and use only 
   `reviewed-by:` to avoid overlap inflation. The `commenter:` approach was 
   an overcorrection from the CodeRabbit review.

3. **Wrong date filter for reviews** — `merged:>=DATE` is incorrect for scoping 
   review activity. Will switch to `created:>=DATE` on the `reviewed-by:` 
   query since GitHub Search scopes `reviewed-by:` + `created:` to when the 
   review was submitted, not when the PR merged.

Working on the fix now.

.github/scripts/fetch-leaderboard.js
yagyaXsen: Fixed in commit 3ef6d08 — removed the `commenter:` query entirely 
and now use only `reviewed-by:` as the single source for the 
`reviews` count.

The `commenter:` approach was added to try to capture more review 
activity, but as you correctly pointed out, a formal GitHub review 
also generates comment activity — meaning the same action was being 
counted in both `prsReviewedCount` and `reviewCommentsCount`, 
inflating the metric.

`reviewed-by:` alone is the cleaner and more accurate signal — it 
counts distinct PRs where the user submitted a formal review, with 
no overlap risk.

.github/scripts/fetch-leaderboard.js
yagyaXsen: Fixed in commit 3ef6d08 — replaced `merged:>=DATE` with `created:>=DATE` 
for the `reviewed-by:` query.

You are right that `merged:` was the wrong qualifier here — it was 
scoping to when the PR was merged rather than when the review was 
actually submitted, which could pull in old review activity from 
outside the selected period.

Switched to a single `sinceFilter` using `created:>=DATE` applied 
consistently across all three activity types (issues, PRs, reviews). 
When used with `reviewed-by:USERNAME`, the `created:` qualifier 
filters by when the review event was created — which correctly scopes 
the count to review activity that actually happened within the 
selected period.

Also cleaned up the two separate filter variables 
(`sinceFilterForIssues` and `sinceFilterForReviews`) into one unified 
`sinceFilter` since the same `created:>=DATE` now applies to all 
queries.

.github/scripts/fetch-leaderboard.js
yagyaXsen: Fixed in latest commit — updated `isRateLimited` to also catch 
GitHub's secondary rate-limit case by checking for the presence 
of a `Retry-After` header on 403 responses:

```javascript
const isRateLimited =
  res.status === 429 ||
  (res.status === 403 && res.headers.get('x-ratelimit-remaining') === '0') ||
  (res.status === 403 && res.headers.get('retry-after') !== null);

**Comment by yagyaXsen** [comment]
@hiyach28 all four points from your review have been addressed:

1. User existence cached once per username — period queries only
   run for confirmed GitHub users 
2. `commenter:` removed — using `reviewed-by:` only to avoid
   double counting
3. `merged:>=DATE` replaced with `created:>=DATE` for accurate
   review period scoping
4.  Secondary rate-limit (403 + Retry-After) now detected and
   retried explicitly

Ready for re-review when you get a chance.

**Review by Maanvi212006 — COMMENTED** [review]
Please attach a screen recording in the description showing the changes.

**Comment by yagyaXsen** [comment]
https://github.com/user-attachments/assets/74c3efd1-4e3d-406b-8b5d-2df2f08f1b4c

@Maanvi212006 The recording below shows the updated Community Leaderboard with 
GitHub activity statistics added via this PR.

**What's shown:**
- Live site (before) — no GitHub columns
- Preview (after) — new Issues Opened, PRs Opened, PR Reviews columns
- Podium cards showing GitHub stats row (Issues | PRs | Reviews)
- Period switching (Weekly / Monthly / All Time)


> Note: GitHub activity columns show 0 in the preview because the 
> Netlify deploy preview does not re-run the nightly data fetch script. 
> In production, the CI workflow runs with `secrets.GITHUB_TOKEN` and 
> will populate real GitHub stats for contributors whose Discourse 
> username matches their GitHub username.

**Comment by yagyaXsen** [comment]
@hiyach28 all requested changes have been addressed . Could you please 
re-review when you get a chance?

**Comment by Maanvi212006** [comment]
@yagyaXsen can you please try running `node .github/scripts/fetch-leaderboard.js` . That will update `_data/leaderboard.json` .

**Review by banana-three-join — CHANGES_REQUESTED** [review]
The data collection itself is heading in the right direction, but there are a few things that need to change before this can go in — mainly around the scoring system, the sorting regression, and a number of unrelated changes that have been bundled in.
 
Grouped by priority below.
 
---
 
## Blocking
 
### 1. The scoring system was modified, and that wasn't part of the request
 
Issue #2917 asks to *collect* GitHub contributor activity and store it "in a machine-readable format that can be consumed by the Community Leaderboard". It does not ask for the scoring mechanism to change.
 
This PR introduces a `github_score` with invented weights (`issues * 2 + prs * 3 + reviews * 2`) and folds it into a new `total_score` that the UI now ranks and displays. The weights aren't justified anywhere, and they're on a completely different scale from the existing Discourse points.
 
The effect is visible in the `_data/leaderboard.json` committed in this PR, all-time period:
 
- master: rank 1 Lee (1391), rank 3 theBeginner86 (182)
- this PR: rank 1 theBeginner86 (1484), rank 2 Lee (1391)
The single person in the hardcoded username mapping displaces the top contributor. This also contradicts the "Existing Discourse leaderboard data is untouched" checkbox in the description.
 
Please drop `github_score` and `total_score`. Keep `score` exactly as it is today, and surface `github_issues`, `github_prs` and `github_reviews` as displayed statistics only.
 
### 2. Sorting regression
 
`buildLeaderboard` in master sorts by score descending, then `username.localeCompare` ascending. The new re-sort in `buildAllPeriods` is `.sort((a, b) => b.total_score - a.total_score)` with no tiebreak, which removes the alphabetical ordering.
 
It's worth being precise about what actually happens: because `Array#sort` is stable, the previous ordering leaks through, so ties now resolve by Discourse score descending first and only then alphabetically. That's an undocumented and non-obvious tiebreak rather than no tiebreak at all, which makes it harder to spot.
 
If the scoring change is reverted per point 1, this re-sort should go away entirely. If any re-sort remains, it needs the explicit `|| a.username.localeCompare(b.username)` tiebreak.
 
### 3. Rate-limit exhaustion silently rewrites the leaderboard
 
@hiyach28 raised this and it hasn't been fixed — the reply only broadened the `isRateLimited` detection, not the fall-through.
 
Once `attempt >= 2`, control reaches `if (!res.ok) throw`, the caller catches it, `githubCache[cacheKey] = null`, and the user is assigned zeros. Those zeros then feed `total_score`, change ranks, and get committed to master by the bot on the next scheduled run. A transient API blip permanently rewrites the leaderboard with no signal that anything went wrong.
 
The `404 -> return null` path in the user lookup has the same problem: it doesn't throw, so the `catch` never fires and nothing is logged.
 
Either fail the run explicitly when stats can't be retrieved, or preserve the previous values rather than writing zeros.
 
### 4. The rate-limit pacing doesn't work at scale
 
The authenticated search API allows 30 requests/minute (documented, and the unauthenticated ceiling is 10/minute). Per user per period the script makes 3 search calls separated by only 120 + 120 ms, with a further 200 ms between iterations — 440 ms of sleep per 3 search requests. Measured round-trip on that endpoint is 40–220 ms, which puts the script somewhere around 150–300 requests/minute, several times over the limit.
 
With one entry in the mapping this never surfaces. With 50 users it's 450 search requests, constant secondary rate limiting, the 3-attempt cap, then the silent zeros described above.
 
Either derive the delay from the documented limit (2s+ between search calls) or replace the three search calls per user with a single GraphQL `contributionsCollection` query.
 
### 5. Deprecated search endpoint
 
The script uses legacy `GET /search/issues` without `advanced_search=true`. GitHub deprecated the legacy behaviour (the notice gives a removal date of 4 Sep 2025) with `advanced_search=true` as the REST replacement. It still returns 200 today, but new code shouldn't be written against it.
 
One syntax note for the migration: under advanced search, a space between multiple `repo`/`org`/`user` qualifiers is treated as AND rather than OR. There's only one `org:` qualifier here, so this is a safe change.
 
---
 
## Should be fixed
 
### 6. Review counts are filtered by the wrong date
 
`fetchPeriodStats` appends `created:>=${since}` to all three queries, including the `reviewed-by:` one. `created` filters on when the PR was opened, not when the review happened, so a review submitted this week on an older PR is not counted.
 
The size of the gap is easy to see against a currently active reviewer. `reviewed-by:leecalcote org:meshery is:pr` returns 158 PRs all-time; adding `created:>=2026-08-06` (the script's monthly window) leaves 32, while 56 PRs in that reviewed set have had activity inside the same window, including ones opened as far back as March. Everything reviewed in the window but opened before it is dropped.
 
Commit `9e894fd` is titled "fix review date filter" but the behaviour is unchanged.
 
### 7. `GITHUB_TOKEN` scope
 
The workflow declares `permissions: contents: write` at the top, so the script now receives a write-capable token to make read-only calls against public data. It doesn't need write at all — the commit step uses the `GH_ACCESS_TOKEN` PAT supplied to `actions/checkout`, not `GITHUB_TOKEN`. Please add a job-level `permissions: contents: read`.
 
Also, `process.env.GH_ACCESS_TOKEN` in the fallback chain is dead in CI, since that variable is never passed into the step's `env`.
 
### 8. ESLint changes aren't needed and the description is inaccurate
 
The description says an ESLint Node override was added for `.github/scripts/**`. `eslint.config.mjs` is unchanged in this PR, and master already contains that exact override:
 
```js
{ files: [".github/scripts/**/*.js"],
  languageOptions: { sourceType: "commonjs", globals: { ...globals.node } } }
```
 
So `/* eslint-env node */` isn't just redundant — it isn't recognised under flat config. Running eslint 9.39 with the repo's own config against this file gives:
 
```
ESLintEnvWarning: /* eslint-env */ comments are no longer recognized when linting
with flat config and will be reported as errors as of v10.0.0.
```
 
Please remove both the `eslint-env` comment and `'use strict'` (the override already sets `sourceType: "commonjs"`).
 
### 9. Hardcoded username mapping
 
`GITHUB_USERNAME_MAPPING` ships with a single hardcoded contributor. That's arbitrary, and it's the same contributor who ends up at rank 1 as a result. It should start empty.
 
More importantly, the design requires a code change and a PR review to onboard every contributor. This belongs in `_data/` as YAML, or should be derived from Discourse's associated-accounts field so it maintains itself.
 
### 10. Zero is the wrong default for unlinked users
 
Every user without a mapping renders `0 / 0 / 0` in the table and podium. In the JSON committed here that's 48 of 49 all-time entries, and all 50 entries in both the weekly and monthly periods — only one row in the entire file has a linked GitHub account. Zero asserts the contributor has no GitHub activity, which is false; it means they aren't linked.
 
Use an em dash, or omit the GitHub cells entirely for unmapped users.
 
### 11. Colours should come from the CSS variables
 
`CONTRIBUTING.md` asks contributors to use the global variables in `_sass/rootvariables.scss` via `var()` and to avoid introducing their own. `#7A848E`, `#ffffff` and `rgba(0, 0, 0, 0.08)` are all new values.
 
`--color-primary-super-dark` is `#818c9a` in both themes, so using it for `.github-link` removes the need for the `body.dark-mode .github-link` override entirely. Hover should use `--link-hover-color` rather than `--brand-color-secondary` directly. `.podium-github-stats`'s `border-top` has the same issue.
 
### 12. Unrelated layout changes in `leaderboard.scss`
 
None of these relate to GitHub statistics, and all of them change existing responsive behaviour:
 
- podium `gap` 3.5rem to 2rem
- `.podium-card` 290px to 220px at all widths
- tablet breakpoint 992px to 768px, mobile 768px to 600px
- `.podium-card` sizing removed from the tablet block
- `margin-top: 2rem` removed from `.podium-username-link`
- `gap: 0.6rem` removed from the member anchor rule
Please revert these here. If they're improvements, they're worth their own PR where they can be reviewed on their merits.
 
The description also claims "Fixed existing CSS bugs (invalid box-shadow/border declarations, grey typed-text in filter)" and "Cleaned redundant dark-mode overrides", but none of that appears in the diff.
 
### 13. `nth-child(n+6)` is brittle and over-matches
 
Columns are now Rank(1) through Solutions(5), Issues(6), PRs(7), Reviews(8), Total Points(9). The selector also re-centers Total Points, an existing column. Please put a class on the new cells instead.
 
### 14. `_data/leaderboard.json` shouldn't be in this PR
 
It's generated nightly by the workflow against master. Hand-committing a regeneration creates conflicts with the bot's commits, and the snapshot here is already stale (`last_updated` 2026-09-05T10:23 versus master's 2026-09-06T01:46). Let the workflow produce it after merge.
 
---
 
## Minor
 
### 15. Wasted API calls in the period loop
 
The nested `for (period of PERIODS) { for (username of allUsernames) }` fetches stats for every user/period pair, including users who don't appear in that period's list at all. Iterate `periods[period]`'s own members instead.
 
### 16. Unnecessary avatar/markup changes
 
In `fetch-leaderboard.js` the `avatar` change is pure reformatting — one line split across three with no behavioural change. Please drop it to keep the diff focused.
 
In `leaderboard.html`, the single anchor is split into `avatar-link` and `username-link` pointing at the same URL, producing two adjacent links to an identical destination. If the restructure is only to make room for the GitHub link, the avatar and username can stay inside one anchor.
 
### 17. Redundant `total_score || score` fallback
 
`total_score` is always `score + github_score`, both non-negative, so if `total_score` exists it's greater than or equal to `score`, and `total_score === 0` implies `score === 0`. The fallback only helps for JSON predating this PR — and this PR commits regenerated JSON, so it's dead on arrival. The `||` chain would also mask a legitimate 0.
 
### 18. Logging
 
`console.log` / `console.warn` do appear in the Actions run log, so this is 9 lines per user per run once the mapping is populated. Worth trimming.
 
The ternary inside the log is also dead code: within the `try`, `fetchPeriodStats` either returns an object or throws, so `: 'not found on GitHub'` is unreachable.
 
### 19. Period windows are approximate and not aligned with Discourse
 
`PERIOD_SINCE` computes the boundary from the script's run time and truncates it with `.slice(0, 10)`. Since `created:>=YYYY-MM-DD` includes the whole of that day, the actual windows are 7–8 days and 30–31 days rather than 7 and 30.
 
They're also defined independently of whatever window Discourse applies to its own `weekly` and `monthly` periods, so there's no guarantee the two halves of a row cover the same span of time. Less critical if the scoring change is reverted, but the columns should still be labelled honestly.
 
### 20. `reviewed-by:` doesn't quite match what was asked for
 
Issue #2917 says "Pull requests reviewed (comments made)". `reviewed-by:` counts distinct PRs with a formal review, so five reviews on one PR count once. That's a defensible proxy, but the column label should reflect what's actually being counted.
 
### 21. Small things
 
- `fetchPeriodStats(username, since, githubUser)` never uses `username`.
- `type:issue` and `is:pr` are used inconsistently across the three queries.
- `encodeURIComponent` is applied to the username but not the rest of `q`, so `created:>=` sends an unencoded `>` in the query string.
- `githubFetch` attaches the token to whatever URL it's handed, with no host check.
- The table now has 9 columns with three set to `white-space: nowrap`, but `.leaderboard-table { min-width: 600px }` was left as-is and nothing is hidden at small widths. Worth deciding whether the GitHub columns should collapse on mobile.

banana-three-join:
Also, there's a DCO issue. Please also look into the commit with the missing signoff @yagyaXsen

yagyaXsen:
Hi @Maanvi212006 @banana-three-join , I’ve addressed the review feedback and updated the PR accordingly.

One issue we identified during implementation was the GitHub profile association. The initial approach assumed that a contributor’s Discourse username was also their GitHub username, which could result in incorrect GitHub profiles and statistics being associated with contributors. This has now been corrected by using explicitly verified Discourse-to-GitHub mappings, with no username guessing for unmapped contributors.

I’ve also removed the generated _data/leaderboard.json, preserved the existing Discourse score and ranking behavior, made GitHub activity display-only, corrected review statistics to use actual review timestamps, and addressed the GitHub API rate-limit and workflow permission concerns.

The PR description has been updated

The PR is now ready for re-review.

Maanvi212006 (changes requested):
Thanks for the update. Most of the things look correct now. A few items from the original comment are still outstanding though:

- The username mapping (`_data/github_mappings.json`) is no longer guessing usernames, which is good, but it still ships with a single hardcoded entry (`theBeginner86`) and still requires a manual file edit + PR to onboard any new contributor. Can this start empty, or better, be sourced automatically from Discourse's associated-accounts field so contributors can self-serve instead of needing a PR each time? (as @banana-three-join suggested)

- In `_sass/leaderboard.scss`, `.podium-github-stats`'s `border-top` still uses raw `rgba(0, 0, 0, 0.08)` / `rgba(255, 255, 255, 0.12)` instead of a variable from `_sass/rootvariables.scss`, same issue as originally flagged.

- The podium gap (2rem to 3.5rem), `.podium-card` width (220px to 290px), and the tablet/mobile breakpoints (768px/600px to 992px/768px) are still changed in this PR and are unrelated to GitHub stats. These are the exact same values already merged to master via #2924, so this PR is just re-carrying an already-shipped change alongside the unrelated bits. Please revert these lines here since they add nothing over what master already has.

- `since` is still interpolated unencoded into the query string (`+created:>=${since}`); low severity since the value is a fixed-format date, but worth wrapping in encodeURIComponent for consistency with the rest of the query.

- Minor/unaddressed from before: PERIOD_SINCE still truncates with .slice(0, 10), so weekly/monthly windows are approximate (7-8 / 30-31 days) and not guaranteed to align with Discourse's own period boundaries. Not blocking, but worth labeling honestly if left as-is.

Also, for the record: the PR description's claim about an ESLint Node override being added doesn't hold up — eslint.config.mjs isn't touched in this branch, and master already has that override. Worth correcting the description so reviewers aren't checking for a change that isn't there.

yagyaXsen:
Hi @Maanvi212006, thanks for the detailed follow-up.

I investigated sourcing the GitHub identity through Discourse's public API, including the directory and user endpoints. Unfortunately, the public responses do not expose associated_accounts, github_username, or another usable GitHub identity field. The user custom_fields are also empty. Automatically exposing that association would require an admin-level API/custom Discourse-side change, which is outside the scope of this PR.

For now, I've retained _data/github_mappings.json as an explicit verified source of truth, with no username guessing or automatic Discourse → GitHub assumptions. Unmapped contributors continue to show —.

I've also addressed the remaining items:

replaced the raw podium border colors with the existing theme variable;
reverted the unrelated podium/layout changes already present in master;
encoded the since query parameter;
documented the approximate GitHub period windows;
corrected the PR description regarding the ESLint override.

git diff --check, node --check, and ESLint are all passing, and the PR diff contains only the intended five files.

The PR is ready for re-review.

yagyaXsen:
Hi @Maanvi212006, regarding the GitHub identity mapping, I was thinking that instead of maintaining _data/github_mappings.json manually, we could address this more systematically through a separate issue.

We could add an option for contributors to link their GitHub account to their Meshery/Discourse profile, either during signup or later through their profile/account settings. This would give us an explicit user-provided association that the leaderboard could use instead of relying on username matching or maintaining mappings manually.

I don't think GitHub should necessarily be mandatory for every user, since contributors may not use GitHub. A "Connect GitHub" / "Link GitHub account" option would probably be a better approach.

If this sounds reasonable, we can create a separate issue for the account-association flow and keep this PR focused on displaying the GitHub activity once a verified mapping is available.

yagyaXsen:
Hi @Maanvi212006, I wanted to discuss the GitHub username mapping part a little further.

Right now, we are using _data/github_mappings.json as a verified mapping, which is safe because we never assume that a Discourse username is the same as a GitHub username. However, the limitation is that whenever a new contributor wants their GitHub activity to appear, someone would have to manually add their GitHub username to this file and create a PR.

I checked the public Discourse API to see if we could automatically get the GitHub account associated with a Discourse user, but the public API doesn't expose the associated_accounts information.

So I was thinking that we could create a separate follow-up issue for solving this properly. Since Discourse already knows which GitHub account is associated with a user when GitHub SSO is used, we could potentially use an authenticated/read-only Discourse API key stored securely as a GitHub Secret. The leaderboard workflow could then ask Discourse for the user's verified GitHub account and use that account to fetch the GitHub activity.

This would mean:

Discourse user → verified GitHub account → GitHub activity

instead of:

Discourse user → manually maintained mapping file → GitHub activity

It would also avoid username guessing and allow new contributors to be handled automatically without requiring a new PR every time.

I think this is better handled as a separate issue because it involves authenticated Discourse API access and is beyond the scope of this PR. For this PR, we can keep the current verified mapping approach.

Would you be okay with creating a follow-up issue for this automated approach?

_data/github_mappings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "theBeginner86": "theBeginner86"
```
hiyach28: Could we initialize this mapping as an empty object instead?

Since this file is intended to hold explicit mappings for cases where a Discourse username doesn't match the GitHub username, having a single contributor pre-populated here seems unnecessary and makes the initial data look contributor-specific.

I think starting with `{}` would keep this neutral, and mappings can then be added only when they're actually needed.

banana-three-join:
Remove the manual mapping. Instead, we'll have two separate boards, same page, just different stats, different rankings, no need for tying the data together between sources.

Because of this, both the issue and the PR need to be updated @Maanvi212006 @yagyaXsen

Maanvi212006:
I have updated the issue and PR description, just one thing to confirm @banana-three-join : there will be a separate json file for github-related data (`_data/github_leaderboard.json`) right?

yagyaXsen:
Thanks @banana-three-join, understood.

I’ll remove the manual Discourse → GitHub mapping and keep the two boards completely independent on the same page.

The existing Community leaderboard will retain its current stats and ranking, while the GitHub leaderboard will use GitHub contributor data directly with its own stats and ranking. There will be no mapping or combining of users between the two sources.

Maanvi212006:
@yagyaXsen any updates? If you have any doubts, please let us know!

banana-three-join:
So future reviewers and under the same context, a conversation was had over in Slack which requested the addition of a second leaderboard which is going to be populated by the Github contributor activity statistics.

The requested second leaderboard is to populate its rows with data completely separately from the `discuss` endpoint.

yagyaXsen:
@Maanvi212006 I understand that the manual Discourse → GitHub mapping should be removed and the two boards should remain completely independent, with different stats and rankings.

yagyaXsen:
@banana-three-join @Maanvi212006
I’ve removed the manual Discourse → GitHub mapping and the combined scoring system. The GitHub board now discovers contributors directly from GitHub and displays Issues Opened, PRs Opened, and PR Reviews independently from the Community leaderboard.

I’ve intentionally not introduced a new scoring formula or metric priority. For now, GitHub contributors are displayed alphabetically.

Could you confirm how you would like the GitHub leaderboard to be ranked — for example, by PRs opened, reviews, issues opened, or another metric?

Maanvi212006:
Thanks much @yagyaXsen ! Regarding the scoring formula, I think we can do this priority order: PRs opened > Issues opened > PRs reviewed. A formula like : `3*PRs opened + 2*Issues opened + 1*PRs reviwed` as we use something similar in the discourse leaderboard: `3*solutions + 2*likes + 1*posts` . 
Thoughts @banana-three-join @hiyach28 ?

Also, could you please attach a screen recording showing the changes in the PR description?

yagyaXsen:
Thanks @Maanvi212006! I've implemented the proposed GitHub activity ranking as:

3 × PRs opened + 2 × Issues opened + 1 × PRs reviewed

GitHub remains a separate board from the existing Community/Discourse leaderboard, with no identity mapping between the two.

I've also attached a short screen recording demonstrating the separate boards, period switching, GitHub activity stats, ranking, and filtering.

@banana-three-join @hiyach28 please confirm if this ranking formula/order looks good.

https://github.com/user-attachments/assets/0022432b-adbe-406e-9a2d-c44d5041e1d2

Maanvi212006:
Yes, looks good! But, for that "Community" and "GitHub" switch, can we use a toggle instead of buttons?

yagyaXsen:
Thanks @Maanvi212006 ! I've updated the Community/GitHub switch to a toggle as suggested and pushed the changes to the PR. Please have a look.
Screenshot 2026-09-14 at 10 33 18 AM

.github/scripts/fetch-leaderboard.js
```diff
@@ -51,41 +62,204 @@ function buildLeaderboard(items) {
...
+    })
+    .map((user, index) => ({
+      ...user,
+      rank: index + 1
+    }));
+
+  return leaderboard.slice(0, 50); // top 50
+}
+
+async function buildAllGitHubPeriods() {
```
Maanvi212006, line 277: Should this also refuse to save if periods.all comes back empty, like the Discourse build does above? Right now a GitHub API hiccup could silently overwrite github_leaderboard.json with empty data.

Maanvi212006 (changes requested):
New layout wrappers (`period-select-wrapper`, `board-tabs-wrapper`, `filter-wrapper`) use inline `style=` for desktop layout instead of SCSS, inconsistent with the rest of the component's styling pattern.

.github/scripts/fetch-leaderboard.js
Maanvi212006: addressed

.github/scripts/fetch-leaderboard.js
```diff
@@ -51,41 +62,207 @@ function buildLeaderboard(items) {
...
+      };
+    }
+  };
+
+  const dateFilter = since ? `+created:>=${encodeURIComponent(since)}` : '';
+  const searchOpts = '&advanced_search=true';
+
+  // 1. Issues
+  let page = 1;
+  while (page <= 5) { // max 500 items to save quota
```
hiyach28: These loops stop at 500 results (here and at line 151) and don't set `sort`, so GitHub returns "best match" order. All Time ends up ranked from an arbitrary sample of about 8.5k issues and 21k PRs, and Monthly is already affected too (547 PRs in the last 30 days). The search API also can't return more than 1,000 results per query. Could we split the queries into date ranges (e.g. per month) so each stays under the limit? Or drop All Time for the GitHub board, or at least log a warning when `total_count` is higher than what we fetched.

.github/scripts/fetch-leaderboard.js
```diff
@@ -51,41 +62,207 @@ function buildLeaderboard(items) {
...
+    if (data.items.length < 100) break;
+    page++;
+    await sleep(2000);
+  }
+
+  // 3. PR Reviews
+  const candidatePrs = [];
+  page = 1;
+  const updatedFilter = since ? `+updated:>=${encodeURIComponent(since)}` : '';
+  while (page <= 2) { // Limit to 200 PRs to keep API usage reasonable
```
hiyach28: Review counts only look at the first 200 PRs. 328 PRs were updated in the last 7 days and 772 in the last 30, so reviews on the rest are dropped without any sign. For All Time there's no date filter at all, so it's 200 arbitrary PRs. Please page through the full set (with `sort=updated`), or show the cap in the UI.

.github/scripts/fetch-leaderboard.js
```diff
@@ -51,41 +62,207 @@ function buildLeaderboard(items) {
...
+    page++;
+    await sleep(2000);
+  }
+
+  for (const pr of candidatePrs) {
+    if (!pr.pull_request || !pr.pull_request.url) continue;
+    const prUrl = pr.pull_request.url;
+
+    if (!prReviewsCache[prUrl]) {
+      prReviewsCache[prUrl] = await githubFetch(`${prUrl}/reviews`);
```
hiyach28: Minor: `/reviews` defaults to 30 results per page and isn't paginated here, so PRs with more than 30 reviews are undercounted. Adding `per_page=100` (plus pagination) would cover it.

hiyach28 (changes requested):
Thanks for reworking this, @yagyaXsen. My earlier feedback is addressed:
- Searches now run per period, and each PR's reviews are cached across periods.
- Reviews come from each PR's `/reviews` and are counted once per reviewer per PR, filtered by `submitted_at`.
- Rate-limit exhaustion now fails the run instead of silently writing zeros.

**On the formula:** 3×PRs + 2×issues + 1×review looks reasonable to me, so no concerns there.

One blocker remains: search results are cut off without any warning, so the rankings are wrong on real data. As of today the org has about 8.5k issues and 21k PRs, with 547 PRs created in the last 30 days alone. Details are inline.

Also worth coordinating with #2919: it adds `yearly` to `PERIODS`, but `PERIOD_SINCE` here has no `yearly` entry, so `since` would come back empty and "Year" on the GitHub board would show all-time data. Whichever PR merges second will also conflict in `_includes/leaderboard.html`.

banana-three-join (changes requested):
The two-board split is the right shape and the workflow permission downgrade is correct. Main things to address are the duplication in the fetch script, a few data-correctness problems in how activity is counted, and some leftovers in the CSS.

**Structure and naming**

The three paginated search loops in `collectGitHubContributors` (lines 133–176) are the same code three times: same page counter, same `if (!data || !data.items...) break`, same `length < 100` break, same sleep. Please extract a single helper — something like `searchAll(query, maxPages)` — so the call sites differ only by the query they pass.

Loops 2 and 3 also issue the same underlying query (`org:meshery is:pr`) with different date qualifiers. The PR pass can feed the review pass instead of querying twice.

`collectGitHubContributors` currently fetches, aggregates, scores, sorts and truncates. Please split the fetching from the aggregation and ranking. As it stands the GitHub ranking is an inline block inside the fetch function while the Discourse equivalent is a named top-level function (`buildDiscourseLeaderboard`), even though the only real differences between them are the score formula and the sort key. Two boards that are conceptually the same should read the same way — either a shared `rankLeaderboard(entries, scoreFn, tiebreak)` or, at minimum, a named `buildGitHubLeaderboard(userStats)` that mirrors the Discourse one.

Related to that: `buildDiscourseLeaderboard` returns its chain directly, while the GitHub side assigns the equivalent chain to a local `leaderboard` and then returns `leaderboard.slice(0, 50)`. The intermediate variable isn't doing any work, so drop it and return the chain — the two boards should follow the same execution path. Once the ranking sits behind a shared helper, the only thing left differing between them is the score function and the sort key.

`githubFetch` has three different contracts in one function: `404` returns `null`, rate-limit recurses, any other non-OK throws. A caller can't tell "no data" from "the call failed", and that ambiguity is exactly what ends up stored in `prReviewsCache`. Pick one contract and make failures explicit.

`process.env.GH_ACCESS_TOKEN` in the fallback chain (line 8) is unreachable in CI. The secret is only used as the `token:` input to `actions/checkout`, which doesn't export an env var, and the script step passes just `GITHUB_TOKEN` — which is always set there and short-circuits the chain anyway. If it's meant as a local-run convenience, worth a comment saying so; otherwise drop it.

`github_leaderboard.json` writes a `default_period`, but the template only ever reads `site.data.leaderboard.default_period` (`leaderboard.html` line 83). Either read the GitHub one or don't write it.

**Data correctness**

The pagination caps mean the monthly and all-time boards don't reflect the data they claim to. Current counts for the org: 21,028 PRs and 8,461 issues all-time, and 546 PRs created inside the monthly window — against a 500-item ceiling (5 pages × 100). Monthly is already truncated today. Raising the cap won't fix it either, since the Search API returns at most 1,000 results per query. If the boards are going to claim all-time totals, the counts need to come from per-repo REST pagination rather than org-wide search; otherwise the periods should be labelled as what they actually are, which is the most recent N items.

The queries don't pass `sort`, and search defaults to best match. So the items you do read aren't the most recent ones, and the ordering can shift between runs for reasons unrelated to activity. If you stay on search, add `&sort=created&order=desc`.

The review candidate set is capped at 200 PRs (2 pages). `org:meshery is:pr updated:>=` returns 261 for the last 7 days and 771 for the last 30, so review counts are already incomplete at the weekly window and roughly a quarter-covered monthly.

`${prUrl}/reviews` (line 183) is fetched without `per_page`, so it returns the first 30 reviews only, with no pagination. Bot reviews count against those 30 before the `type === 'User'` filter runs, so on PRs with heavy automated review activity the human reviews can fall off the page entirely. Add `per_page=100` and page through.

Two opposite failure modes in the save path:

- Discourse data is written first, then any GitHub error throws and hits `process.exit(1)`. Steps default to `if: success()`, so the commit step is skipped and the Discourse leaderboard doesn't get committed either. A GitHub hiccup now stops the Discourse board from updating.
- Conversely, `buildAllGitHubPeriods` has no `if (!periods.all.length)` guard like `buildAllDiscoursePeriods` does (line 71), so an empty-but-successful result silently overwrites the file. This is Maanvi's open comment and it's still outstanding.

`prReviewsCache[prUrl] = null` on a 404 fails the `if (!prReviewsCache[prUrl])` check, so those PRs are re-fetched on every period.

`type === 'User'` doesn't exclude machine accounts. `l5io` is a `User`-type account with 226 PRs in the org and will rank. Please add an explicit denylist.

PRs count at 3 points regardless of whether they were merged or closed unmerged, which is the easiest thing on the board to game. Worth considering crediting merged PRs instead.

`PERIOD_SINCE` truncates with `.slice(0, 10)` and `created:>=YYYY-MM-DD` includes the whole of that day, so the windows are 7–8 and 30–31 days. Fine to leave, but the labels should say "last 7 days" rather than "This Week".

`encodeURIComponent` is applied to `since` but not to the `>=` in front of it.

`type:issue` and `is:pr` are mixed across the queries. Both forms return identical counts under `advanced_search=true`, so this is purely consistency — just pick one.

**Template and styles**

`_data/github_leaderboard.json` isn't in the branch, so the deploy preview can't show the GitHub board and reviewers can't check it without running the script locally. Please confirm the page renders cleanly with the file absent, and consider committing a seed file.

The three new wrappers (`period-select-wrapper`, `board-tabs-wrapper`, `filter-wrapper`) carry inline `style=` attributes. That's why the 768px block needs `!important` on `flex` and `justify-content` — it's overriding your own inline styles. Move the layout into the SCSS and the `!important` goes away.

`#ffffff`, `#000000` and `rgba(0, 179, 159, 0.1)` are raw values. `--color-white`, `--text-color-primary-fixed` and `--brand-color-secondary-50` all exist in `rootvariables.scss`, and CONTRIBUTING.md (lines 56–58) asks contributors to use them rather than introduce new ones.

The `.leaderboard-section` → `.leaderboard` rename leaves no stale references, but `.period-toggle` (`leaderboard.scss` line 406) is now dead CSS since the markup no longer emits that element. The 768px block also dropped `height: 42px` and `box-sizing` from `.leaderboard-filter`, which is an unrelated responsive change — please restore it or split it out.

`leaderboard.md` still describes the page as ranking forum posts, likes and solutions, with no mention of the GitHub board.

The board switch uses `role="group"` with `aria-pressed` buttons. With two mutually exclusive panels, `role="tablist"` / `tab` / `tabpanel` with `aria-controls` and arrow-key navigation is the more appropriate pattern.

Maanvi212006:
something  I found while testing: right now the yearly and all boards return identical data.

Turns out it's because both queries exceed the Search API's 1,000-item cap. Since results are sorted newest-first and capped at 1,000, "most recent merged PRs in the last year" and "most recent merged PRs ever" end up being the exact same 1,000 items once yearly volume alone crosses that ceiling; the since filter never actually gets to exclude anything.

Bumping the page cap won't fix it since 1,000 is a hard limit on the Search API itself, not something we control. The real fix would be per-repo REST pagination instead of org-wide search for these wider windows. We can scope that as a follow-up if we don't want to block this PR on it, but wanted to flag it now since it explains the odd behavior.

Maanvi212006:
Re: merging queries 2 & 3 into one - decided to keep them separate.

I initially explored folding the merge-credit query into the review-candidate query (`updated:>=since`) to cut down on Search API calls, but merge date and update date aren't reliably interchangeable here - a PR could be merged in-period but not show up as "updated" in that same window depending on how GitHub timestamps that action, and vice versa for other activity. Querying `merged:>=since` directly is the accurate way to answer "which PRs were merged in this period," so I think we should rather keep the extra API call than risk silently miscounting merge credit.

Happy to revisit if the extra query is a real concern for rate limits, but just wanted to flag the reasoning.

yagyaXsen:
Thanks for the review feedback. I’ve addressed the outstanding implementation issues in the latest commit 4eff63d.

Key updates include:

Replaced GitHub Search API pagination with per-repository REST pagination to avoid the 1,000-result limit.
Updated PR activity to count merged PRs using merged_at.
Added paginated PR review fetching with submitted_at filtering and reviewer deduplication.
Added rate-limit protection and explicit failure handling.
Added machine-account filtering.
Added yearly period support and kept GitHub and Discourse leaderboards independent.
Updated the related styling and documentation.

DCO, lint, syntax validation, and deploy preview checks are passing.

banana-three-join (changes requested):
Changes needed before this can be approved:

Merging this reverts recent master work without raising a conflict. The branch's earlier merge commits resolved these hunks in its own favour, so the merge is clean and the branch's older values win. Reverted in `_sass/leaderboard.scss`: podium gap 2rem → 3.5rem, `.podium-card` 220px → 290px, tablet breakpoint 768px → 992px, mobile 600px → 768px, `margin-top: 2rem` re-added to `.podium-username-link`, mobile podium margin 5.5rem → 3.5rem, controls gap 0.6rem → 1rem. In `_includes/leaderboard.html` it deletes the `updatePeriodDescription(period);` call added to master on 15 Sep in `4104916`, leaving the function defined with zero call sites so `#leaderboard-period-phrase` no longer updates on period change. These need re-resolving against master explicitly; another merge won't do it.

The run can't finish inside the rate limit. The comments and `remainingRateLimit = 5000` assume 5,000 req/hr, but `GITHUB_TOKEN` in Actions is 1,000/hr per repository. Current volume is around 1,060 calls — roughly 300 pages to walk 29,535 issues and PRs across 16 repos, plus 760 review calls for PRs updated in the last 30 days — and it grows weekly. This needs GraphQL, a PAT, or a much smaller candidate window.

The headroom guard also misses most of those calls. `remainingRateLimit < 200` is only checked in the review loop; `fetchRepositories` and `fetchRepositoryActivity` are unguarded, and the counter starts at 5000 so it's wrong until the first response. There's no delay between the 760 review calls either — `sleep(150)` only runs when a single PR has more than 100 reviews.

When the GitHub half fails, the Discourse update is discarded with it. `main()` writes `leaderboard.json`, then exits 1, and the commit step has no `if:` so it defaults to `success()` and never runs. Use `if: always()`, or only exit non-zero when Discourse failed.

The committed `github_leaderboard.json` isn't this code's output. It has no `default_period` even though `saveJSON` now writes one, and `yearly` and `all` are identical with all-time totals roughly ten times too low. Regenerate it from the current script.

Yearly and all-time reviews aren't "PRs Reviewed". Candidates are PRs updated in the last 30 days, so for those periods the column counts only reviews on recently-touched PRs, and it feeds `github_score`. Either limit the GitHub board to weekly/monthly, drop reviews from the longer periods' score, or relabel the column.

`leaderboard.json` should be renamed `discuss_leaderboard.json`. With a sibling `github_leaderboard.json` the bare name stops meaning anything, and `_data/discuss/meshery.json` already sets the convention. It touches four references to `site.data.leaderboard`, the `saveJSON` filename and the workflow `file_pattern`.

An empty period combined with a filter query renders nothing. `renderLeaderboard` returns early without clearing the tbody or podium, so stale rows stay in the DOM; a later `input` event counts them, `visibleCount > 0`, and the final `else` hides the empty state while the board panel is still `is-hidden`. Clear both tbodies and podiums before returning.

`renderGithubTable` and `renderGithubPodium` should collapse into the Discourse ones. Each pair is around 60% identical and everything that differs is field names and three stat labels. One renderer taking a board config would also remove the `currentBoard === 'discourse' ? ... : ...` ternaries in `renderLeaderboard` and `filterLeaderboard`; right now adding a third board means editing five places.

`.leaderboard-filter:focus-visible` should keep its original value. It went from `rgba(0, 179, 159, 0.15)` to `var(--brand-color-secondary-50)`, which is `rgba(0, 179, 159, .5)`. Every other colour swap in the diff is value-identical; this one triples the focus ring's opacity.

Maanvi212006:
Switched GitHub data fetching from REST to GraphQL.

Why: REST needed one call per PR just to get its reviews, plus REST refuses to paginate past 10,000 results (we hit this on meshery/meshery). Both problems meant yearly/all-time review counts were incomplete or missing.

What changed: One GraphQL query per repo now pulls issues, PRs, and their reviews together, with pagination instead of a hard cap. This gets full history for every period, so the earlier "Reviews: —" workaround for yearly/all-time is no longer needed.

Known limit: Each PR returns up to 30 reviews. A handful of very old, heavily-reviewed PRs (mostly in meshery/meshery and schemas) have more than that — the script logs these and only counts the first 30. Affects <80 PRs out of ~21k, so the impact on scores is minor. Happy to paginate reviews per-PR in a follow-up if we want it exact.

Updated the PR description with the latest screen recording.

banana-three-join (changes requested):
`leaderboard.md` still reads `site.data.leaderboard.default_period`, but the file is now `_data/discuss_leaderboard.json`. The `{% case %}` gets nil and always falls through to `{% else %}`, so the period phrase renders "over the past 7 days" regardless of what `default_period` is set to. It only looks right today because the default happens to be `weekly`. Point it at `site.data.discuss_leaderboard.default_period`.
 
The GraphQL budget assumption is wrong. The comment above `remainingGraphQLPoints` says 5,000 points/hr for both token types, but GitHub's GraphQL docs give `GITHUB_TOKEN` 1,000 points/hr per repository — the same split as REST. The counter also seeds at 5000, while the REST one was deliberately seeded low for safety. A full run is roughly 535 requests (21,050 PRs at 50 per page, 8,485 issues at 75 per page), and the `< 200` guard leaves 800 usable points, so on `GITHUB_TOKEN` anything averaging over about 1.5 points per request fails the run. The queries already select `rateLimit { cost }` but nothing reads it — logging it would settle whether this can run on `GITHUB_TOKEN` at all, or whether `GH_ACCESS_TOKEN` is now a hard requirement worth documenting.
 
Lastly, the `leaderboard.md` copy describes only the Discourse board and, as this is close to being mergeable, I believe the creation of the issue for its addition is good to go.

banana-three-join:
Also, once approved, it would be good to squash it into one commit so 22 commits on a new feature don't get added into the commit history @yagyaXsen @Maanvi212006

yagyaXsen:
Thanks for the detailed review @banana-three-join. I've addressed the requested changes in commit `b6f2f85`:

- **Restored Master SCSS Values**: Restored all recent master properties in `_sass/leaderboard.scss` (podium gap `3.5rem`, card width `290px`, username `margin-top: 2rem`, controls gap `1rem`, tablet breakpoint `992px`, mobile breakpoint `768px`, mobile podium margin `3.5rem`).
- **Restored `updatePeriodDescription()` Sync**: Ensured `updatePeriodDescription(currentPeriod)` runs on every render in `_includes/leaderboard.html`, keeping `#leaderboard-period-phrase` in sync across tab and period changes.
- **GraphQL Rate-Limit Budget Handling**: Updated initial GraphQL headroom floor to a conservative `1000` points and verified synchronization directly against GitHub's server-side `rateLimit.remaining`. Retries with backoff handle HTTP 429/403 and GraphQL `RATE_LIMITED`.
- **Discourse Failure Isolation**: Verified Discourse updates execute and save first; GitHub failures do not prevent Discourse data from being committed, and the workflow commit step uses `if: always()`.
- **Corrected Data References**: Replaced the stale `site.data.leaderboard` reference with `site.data.discuss_leaderboard` in `leaderboard.md`.
- **Updated Documentation**: Updated `leaderboard.md` to describe both the Community (1 pt/post, 2 pts/like, 3 pts/solution) and GitHub (3 pts/merged PR, 2 pts/issue opened, 1 pt/review) leaderboards with independent rankings and rolling periods.
- **Restored Focus-Visible Styling**: Preserved exact `rgba(0, 179, 159, 0.15)` focus rings on filter and period select inputs.
- **Consolidated Renderers & DOM Cleanliness**: Maintained the unified `BOARDS` config and verified zero-activity and empty-search states cleanly clear DOM elements.

**Validation Status**:
- `node -c .github/scripts/fetch-leaderboard.js` (pass)
- `npx eslint .github/scripts/fetch-leaderboard.js` (pass, 0 errors/warnings)
- `git diff --check` (pass, 0 whitespace issues)
- GitHub leaderboard mock tests covering bot filtering, 10-scenario period classification, rate-limit guards, deduplication, and tie-breaking (pass)

**Documented Limitation**:
PR reviews are retrieved via `reviews(first: 30)`. For historical PRs with >30 reviews, reviews beyond the 30th are truncated and a warning is logged. Because reviews are deduplicated to at most 1 point per reviewer per PR, distinct reviewer totals are rarely impacted. If maintainers require exact nested pagination for these historical PRs, I am happy to add it.

Per your review instruction, I have kept the individual commits intact for review clarity. Once approved, I will squash the branch into a single clean commit with DCO signoff.

banana-three-join:
Thanks for the fixes! Two things though. @yagyaXsen 

The scoring explanation in `leaderboard.md` wasn't meant to go in here - the plan was to track it in a separate issue, so please revert that paragraph and we'll open the issue for it.

And `b6f2f85` undoes the `_sass/leaderboard.scss` fixes from `3638237`/`211eac1`. Against master it puts back `.podium-card` at 290px, `.leaderboard-podium` gap at 3.5rem, the 992px and 768px breakpoints, `margin-top: 2rem` on `.podium-username-link`, and the 3.5rem mobile podium margin, and drops `gap: 0.5rem`. The merge is clean so nothing flags it - master's values just lose silently. Worth diffing this file against master before the squash, since it's regressed twice now.

Also, for the next reply since it's a very easy change, there's no need to provide me with the `validation status`, `documented limitation` and any extremely punctual over-explanation of the changes.

yagyaXsen:
Thanks, fixed both. I reverted the GitHub scoring explanation from leaderboard.md and re-aligned _sass/leaderboard.scss with the current master values, including the 0.5rem gap.

banana-three-join:
Ty! @yagyaXsen. Could you update your branch to the latest commit and make sure that everything is now fine? @Maanvi212006. I went over it and it looks good to merge, but another set of eyes would be good on this. After Maanvi's confirmation, you can proceed to squash it so I can merge it @yagyaXsen.

.github/workflows/update-leaderboard.yml
```diff
@@ -32,11 +32,15 @@ jobs:
 
       - name: Fetch leaderboard data
         run: node .github/scripts/fetch-leaderboard.js
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_ACCESS_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
```
Maanvi212006: The PR description says the script receives `GITHUB_TOKEN` (read-only), but the script resolves `GH_ACCESS_TOKEN || GITHUB_TOKEN` (fetch-leaderboard.js:13), so in CI it actually runs with the PAT. Please update the PR description accordingly.

.github/scripts/fetch-leaderboard.js
```diff
@@ -51,45 +81,429 @@ function buildLeaderboard(items) {
...
+        created_at: pr.createdAt,
+        updated_at: pr.updatedAt,
+        merged_at: pr.mergedAt || null
+      });
+      if (pr.reviews.totalCount > pr.reviews.nodes.length) {
+        console.warn(`'${repoName}' PR #${pr.number} has ${pr.reviews.totalCount} reviews, only first ${pr.reviews.nodes.length} counted.`);
+      }
+      for (const r of pr.reviews.nodes) {
+        if (isRealGraphQLUser(r.author) && r.state !== 'PENDING') {
+          reviews.push({ prNumber: pr.number, user: r.author, submitted_at: r.submittedAt, state: r.state });
```
Maanvi212006: Small question on the review dedupe: the review record stores prNumber but not the repo, and the key in `aggregateGitHubPeriods` is `prNumber + ':' + login`. Since PR numbers restart per repo, could two different PRs with the same number in different repos (e.g. #100 in two repos) end up counted as one review for the same reviewer? If so, adding the repo name to the record and to the key might be enough

_sass/leaderboard.scss
```diff
@@ -65,21 +88,58 @@
...
+    justify-content: center;
+    align-items: center;
+    background: transparent;
+    border: 2px solid var(--brand-color-secondary);
+    border-radius: 8px;
+    overflow: hidden;
+    padding: 0;
+}
+
+.tab-button {
```
Maanvi212006: `leaderboard.scss` is imported site-wide via `screen.scss`, and `_includes/partials/catalog-feature.html` also uses a "tab-button" class on its cards. Since `.tab-button` here is unscoped (including the block in the `max-width: 600px` media query), it affects the card on [catalog page](https://meshery.io/catalog/) too. scoping to `.board-tabs` `.tab-button` might be a simple fix.

Preview

Image

Original

Image

.github/scripts/fetch-leaderboard.js
Maanvi212006, line 1: Another thought on reviews: should the PR author also be compared with the reviewer? My understanding is that GitHub can record an author's inline replies on their own PR as a COMMENTED review, which might add to their "PRs Reviewed" count. Would it make sense to store `pr.author.login` on the review and skip when it matches the reviewer? Worth double-checking on a PR where the author replied inline, in case this isn't an issue in practice.

Maanvi212006 (changes requested):
PR description needs updating:

- The script description says activity is fetched "for each contributor via the REST API". but, REST is only used for repo discovery, and issues/PRs/reviews come from GraphQL across the org's repos.

- The workflow now passes both GITHUB_TOKEN and GH_ACCESS_TOKEN, and the script prefers the PAT. Could be worth mentioning, along with the contents: write -> read change.

- The file list has 5 entries but the PR changes 7: the leaderboard.json -> discuss_leaderboard.json rename and leaderboard.md aren't listed.

yagyaXsen:
Thanks for the review! Addressed all three points:

review deduplication now includes repository + PR number + reviewer
leaderboard tab styles are scoped under .board-tabs
PR author self-reviews are excluded

I also updated the PR description with the corrected GraphQL/REST architecture, token permissions, and complete file list. Could you please take another look?

PR #2930, opened 2026-08-31
fix: deterministic tie-breaking for leaderboard scores

### LINKED ISSUE #2929: [UI] Leaderboard: tied scores need a deterministic tiebreaker
## Current Behavior

In `.github/scripts/fetch-leaderboard.js`, `buildLeaderboard()` sorts users by score only:
`.sort((a, b) => b.score - a.score)` When two or more users have the same score, this comparator returns 0 for them, so their relative order falls through to whatever order they arrived in from the Discourse API response. That arrival order isn't based on username, activity, or account creation; it's determined by which `directory_items` row was inserted first into Discourse's internal per-period stats table. This
order is stable across requests (so the leaderboard doesn't visibly "shuffle" on reload), but it has no meaningful relationship to the
user's actual contributions. Two users with identical stats can end up in a fixed, effectively arbitrary rank order relative to each
other, and there's no way to tell from the leaderboard why one outranks the other. 

## Expected Behavior

Ties should be broken deterministically using a signal from the data itself, not from Discourse's internal storage order:

* If users are tied on score, rank them alphabetically by username.

**Note for reviewers**
This fix only affects ordering within the top 50 users per period already returned by the Discourse API (per the `limit=50` in `fetchUsers`). It doesn't address the separate follow-up item (mentioned in #2767) about paginating to score the full member list.

Maanvi212006:
This PR fixes #2929 

## Description
Updated the `.sort()` comparator to break ties deterministically:

* Primary sort: score, descending (unchanged).
* If users are tied on score, sort alphabetically by username as a stable, deterministic tiebreaker.

**Notes for Reviewers**
No other part of the script changes `computeScore`, the mapping steps, `buildAllPeriods`, `saveJSON` and main are all untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved leaderboard ranking consistency when entries have equal scores.
  * Equal-score entries are now ordered alphabetically by username.
  * Score calculations remain unchanged.

* **Updates**
  * Refreshed leaderboard data and its last-updated timestamp.
  * Updated all-time leaderboard rankings to reflect the revised sorting behavior.

### DIFF
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -24,8 +24,8 @@ async function fetchUsers(period) {
 
 function computeScore(item) {
   return (item.post_count || 0) +
-        ((item.likes_received || 0) * 2) +
-        ((item.solutions || 0) * 3);
+    ((item.likes_received || 0) * 2) +
+    ((item.solutions || 0) * 3);
 }
 
 function buildLeaderboard(items) {
@@ -41,7 +41,26 @@ function buildLeaderboard(items) {
       solutions: item.solutions || 0,
       score: computeScore(item)
     }))
-    .sort((a, b) => b.score - a.score)
+    .sort((a, b) => {
+      // Primary: score, descending
+      if (b.score !== a.score) return b.score - a.score;
+
+      // Tied on score. If everyone here has zero score, there's no
+      // activity signal to break the tie with — go alphabetical.
+      if (a.score === 0 && b.score === 0) {
+        return a.username.localeCompare(b.username);
+      }
+
+      // Otherwise break the tie using the same priority the scoring
+      // formula itself uses: solutions (x3) > likes (x2) > posts (x1).
+      if (b.solutions !== a.solutions) return b.solutions - a.solutions;
+      if (b.likes !== a.likes) return b.likes - a.likes;
+      if (b.posts !== a.posts) return b.posts - a.posts;
+
+      // Still fully tied — fall back to alphabetical so the order is
+      // deterministic no matter what order Discourse returned rows in.
+      return a.username.localeCompare(b.username);
+    })
     .map((user, index) => ({
       ...user,
       rank: index + 1
--- a/.github/scripts/fetch-leaderboard.js
+++ b/.github/scripts/fetch-leaderboard.js
@@ -42,23 +42,7 @@ function buildLeaderboard(items) {
       score: computeScore(item)
     }))
     .sort((a, b) => {
-      // Primary: score, descending
       if (b.score !== a.score) return b.score - a.score;
-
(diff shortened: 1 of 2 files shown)


Maanvi212006:
Not ready for review yet. Need to attach the screen recording.

.github/scripts/fetch-leaderboard.js
```diff
@@ -41,7 +41,26 @@ function buildLeaderboard(items) {
       solutions: item.solutions || 0,
       score: computeScore(item)
     }))
-    .sort((a, b) => b.score - a.score)
+    .sort((a, b) => {
```
banana-three-join, line 44: I believe this can be simplified to:

```
if (b.score !== a.score) return b.score - a.score;
return a.username.localeCompare(b.username);
```

computeScore already folds the weighting of each property into a single final score. Re-applying that same priority order (solutions → likes → posts) in the comparator adds the individual properties back into the mix after the score has been computed, which isn't this function's responsibility and defeats the purpose of computeScore.

The `a.score === 0 && b.score === 0` branch is also unnecessary: since the score is the weighted sum of all three properties and none of them can be negative, a score of 0 means all of them are 0. The comparisons below it would return 0 and fall through to localeCompare anyway.

banana-three-join (changes requested):

.github/scripts/fetch-leaderboard.js
Maanvi212006: Yes, I will remove the zero-score branch. For the tiebreak though, just to mention, I decided to use a priority order to mirror the formula (3 * solutions + 2 * likes + posts). However, if using the alphabetical order for non-zero scores too is preferred, then I would simplify this and also update the PR and issue description.

banana-three-join: It's better to just keep the alphabetical order for the other cases as well. If any changes need to be done to how the users are being presented regarding the score, then that's within the responsability solely of the build score function. 

Imagine we need to change the behavior in how the score gets calculated, maybe we need to change the weight of each property but now, there's some external factor in the sorting which changes the behavior so now it doesn't matter if the weights get modified within the function because there's something outside of the function which is affecting it. The secondary sorting name wise just serves as a convention to keep the username list in the same order so it doesn't give the impression that there randomness in the calculation but the `calculateScore` should be the sole source of truth for the ranking.

Maanvi212006: Alright @banana-three-join , makes sense! I will push the requested changes. Also, will update the issue and PR description

banana-three-join: Ty. Let me know so I can approve and merge

banana-three-join (approved):
LGTM. Ty!

Issue #2936, opened 2026-09-05
[Bug]: TypeError: Cannot read properties of null on /catalog/models/


suryaff733:
### **Description:**

**Description**  
Visiting [`/catalog/models/`](https://meshery.io/catalog/models/) crashes with an uncaught `TypeError` in the console on page load, breaking catalog filtering.

**Steps to Reproduce**  
1. Open DevTools (`F12`) → **Console**.  
2. Navigate to `https://meshery.io/catalog/models/`.  
3. Observe the crash:  
   `Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')`

**Root Cause**  
In [`_includes/patterns-filter.html`](file:///Users/surya/My%20Projections/mesheryio/meshery.io/_includes/patterns-filter.html#L270), `compatibilityLink` is `null` because `show_technology="false"` on the Models page. The script calls `.addEventListener()` without checking if the element exists.

**Screenshots**  

<img width="1633" height="768" alt="Image" src="https://github.com/user-attachments/assets/5fabb9f3-e41d-4c7c-8332-eb4c2da7ad66" />

**Suggested Fix**  
```javascript
if (compatibilityLink) {
  compatibilityLink.addEventListener("click", showCompatibility);
}
```

suryaff733:
I would like to work on it.

PR #2938, opened 2026-09-05
[Catalog] Fix null reference TypeError on models page filters

suryaff733:
### Description
This PR fixes a `TypeError` that was crashing the JavaScript runtime on the `/catalog/models/` page.

When `patterns-filter.html` is included with `show_technology="false"` (as is the case for the Models page), the `compatibility-li` element is not rendered in the DOM. However, the script was unconditionally attempting to attach an event listener to it, resulting in `Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')`. This unhandled exception broke catalog filtering and pagination on that page.

This fix adds null checks for both the Category and Compatibility toggle links before attaching their respective event listeners, ensuring the script degrades gracefully when those elements are intentionally omitted by the layout.

**Fixes:** #2936

### Expected Behavior
- The `/catalog/models/` page loads without any JavaScript errors in the console.
- Interactive catalog elements (filters, search, pagination) function correctly on pages where certain filter sections are hidden.

### Screenshots
image

### Environment:
 - OS: All
 - Browser: All
 - Device: All

## Summary by CodeRabbit

* **New Features**
  * Added pattern-type filtering for pattern cards.
  * Added more reliable dropdown interactions across pages with different filter configurations.

* **Bug Fixes**
  * Prevented errors when optional filter elements, such as Technologies, are unavailable.
  * Improved filtering, pagination, visibility updates, and empty-state messaging when content is missing.
  * Filters now start in a consistent cleared state when the page loads.


PARTH-TUSSLE (changes requested):
Hey @suryaff733 , showCategory can still throw when Technologies are hidden
The null check fixes the initial page-load crash, but on /catalog/models/, comp_p and comp_div are also null because the Technologies section isn't rendered when show_technology="false".
showCategory() still accesses both unconditionally, so clicking Categories can still throw a TypeError.
Could you guard those accesses as well and add a small regression test for the show_technology="false" case? This should make sure the Models page is fully functional, and not just fix the initial load error.

suryaff733:
hey @PARTH-TUSSLE my bad fr, you were right 💀

fixed showCategory(), showCompatibility() and showTechnology() — all three were hitting comp_p and comp_div without checking if they even exist (they don't on the models page lol). wrapped everything in null guards now so it won't blow up anymore

also guarded container, prevButton, nextButton and the #not-found element while i was at it — no more unchecked dom access anywhere

should be solid now, lmk if anything else looks off

PARTH-TUSSLE:
> hey @PARTH-TUSSLE my bad fr, you were right 💀
> 
> fixed showCategory(), showCompatibility() and showTechnology() — all three were hitting comp_p and comp_div without checking if they even exist (they don't on the models page lol). wrapped everything in null guards now so it won't blow up anymore
> 
> also guarded container, prevButton, nextButton and the #not-found element while i was at it — no more unchecked dom access anywhere
> 
> should be solid now, lmk if anything else looks off

Looks good, thanks for fixing those remaining DOM accesses👍
One thing that's still missing is the regression coverage for the show_technology="false" case. Could you add a small test/verification for the Models page so this regression doesn't comeup again?

suryaff733:
Hey @PARTH-TUSSLE, here is the verification recording/screenshot of the Models page (show_technology="false"). As shown, clicking the category filters works cleanly with zero TypeErrors in the console.

https://github.com/user-attachments/assets/23618219-a81f-4e46-8bca-ccbdc58455a8

PARTH-TUSSLE (changes requested):
> Hey @PARTH-TUSSLE, here is the verification recording/screenshot of the Models page (show_technology="false"). As shown, clicking the category filters works cleanly with zero TypeErrors in the console.
> 
>  pr2938.mp4

The null-safety fixes looks good, and the verification video confirms the Models category filter no longer throws.
Could you keep this PR scoped to the fix tho? The sessionStorage filter persistence and removal of the pattern-type filtering change behavior unrelated to the null-reference issue. I'd suggest reverting those parts and keeping the PR focused on the DOM null guards.

suryaff733:
Makes total sense, thanks for the review @PARTH-TUSSLE!

I'll revert the sessionStorage persistence and restore the patternType filtering logic so this PR stays strictly scoped to the DOM null-safety guards.

Pushing an update shortly!

suryaff733:
Thanks for the feedback @PARTH-TUSSLE!

I've reverted the `sessionStorage` filter persistence and restored the `patternType` filtering logic. The PR is now strictly scoped to the DOM null-safety guards for cases where `show_technology="false"`.

Katotodan (commented):
Look good overall, two observations and suggestions.
1. Could you use the Optional chaining to save the number of line?
E.g. 
`categoryLink?.addEventListener("click", showCategory); compatibilityLink?.addEventListener("click", showCompatibility); `
through out your changes?
2. Renaming the variable to use camelCase which is the conventional choice in JavaScript is a good practice, could we also shift from var to let at the same time?
As you are doing this improvement.

suryaff733:
Thanks @Katotodan! 

Updated the listeners to use optional chaining (`?.`) and shifted `var` to `let`. Pushed the changes in the latest commit!

PARTH-TUSSLE (approved):
No further changes from my side. LGTM 👍

PARTH-TUSSLE:
Hey could you add this as an agenda item for websites meet on Monday, we could discuss it there, you can find the link to the meeting minutes in the #websites channel on Slack.

Issue #2957, opened 2026-09-10
[Docs] Create VISION.md and vision artifacts for meshery.io


akkki007:
### Current State:

`meshery/meshery.io` has `GOVERNANCE.md`, `MAINTAINERS.md`, `OWNERS`, `CONTRIBUTING.md`, and `CODE_OF_CONDUCT.md`, all of which describe *who decides* and *how to contribute*. None of them states what the website itself is for, or what belongs on it versus what belongs on docs.meshery.io, in the Meshery application, or in `meshery/meshery` itself.

That boundary is currently carried in reviewers' heads. The repository holds 394 model pages under `collections/_models`, 357 catalog pages under `collections/_catalog`, 43 blog posts, 25 extensions, and generated data in `_data/` that is written by `update-catalog.yml`, `delete-catalog.yml`, `update-leaderboard.yml`, `discussions-data-files-update.yml`, and `error-code-updater.yml`. A contributor cannot tell from the repository alone which of those trees are hand-authored and which are synchronized from upstream, so hand edits land in generated pages and get silently overwritten on the next workflow run.

No `VISION.md` exists anywhere in the `meshery` organization today.

### Desired State:

A `VISION.md` at the root of `meshery/meshery.io` that states the site's purpose, its governing principles, and its explicit non-goals, plus the supporting artifacts that make it verifiable and durable. Follows the pattern established for the Sistent design system in layer5io/sistent#1830, adapted to what this repository actually is: a Jekyll site, not a component library.

**Deliverables**

1. `.claude/skills/vision/SKILL.md` - a reusable `/vision` skill, sitting alongside the existing `.claude/skills/meshery-blog-writer/` skill and following its conventions. The skill derives principles from the repository rather than templating them, and it verifies every piece of evidence before writing it down.
2. `VISION.md` at the repository root.
3. `docs/vision/vision-evidence.md` - a claim-by-claim evidence sheet. Every sentence in `VISION.md` appears here with a resolvable citation: a path that exists on `master`, a PR or issue number that resolves, or a published URL that contains the quoted text. Claims that describe intended direction rather than shipped state are marked as such.
4. `docs/vision/vision-hypotheticals.md` - ten stress-test hypotheticals with maintainer verdicts and reasoning quoted verbatim, and a changelog mapping each verdict to the edit it produced in `VISION.md`.

**Scope and constraints**

- Documentation only. No changes to Jekyll configuration, layouts, includes, plugins, workflows, or any published page.
[...]

PR #2958, opened 2026-09-10
docs: add VISION.md

akkki007:
## Description

Adds `VISION.md` at the repository root. Part of #2957.

Scoped to that one file per review. The earlier revision of this branch also carried an evidence sheet, a calibration record, and a reusable skill; those are held back for follow-up pull requests so this one stays reviewable.

## What it contains

Purpose, goals, five guiding principles, boundaries and non-goals, and success criteria, following the structure suggested in review.

The five principles:

- **Public web presence.** The site is judged as a website: does a visitor understand Meshery better and know where to go next.
- **Present, don't own.** Models, integrations, catalog entries, and community activity originate elsewhere and are published here. Wrong data is fixed at its source; presentation is what this repository owns.
- **Generated content is generated.** Improve the systems that produce it rather than editing output by hand.
- **Documentation belongs to docs.meshery.io.** This site says what exists and why it matters; instructions live in the documentation and meshery.io links to them.
- **Collections define published content.** New content types are deliberate additions to the information architecture, not one-off pages.

## Altitude

Per review, this is written as a durable north star rather than a description of the current implementation. It contains no page counts, no workflow names or behavior, no file-level regeneration mechanics, and no `_config.yml` settings. Those facts belong in architecture or contributor documentation, and they change over time.

The principles are still derived from how the repository actually works. That derivation and its evidence move to a follow-up rather than into this document.

## Scope

Documentation only. One new file, no changes to Jekyll configuration, layouts, includes, plugins, workflows, or any published page. `VISION.md` is not added to the `_config.yml` `exclude` list, matching how `CONTRIBUTING.md`, `GOVERNANCE.md`, and `SECURITY.md` are already handled.

## Summary by CodeRabbit

* **Documentation**
  * Rewrote the project vision statement to clarify the site’s purpose, goals, guiding principles, boundaries, non-goals, and success criteria.
  * Replaced prior operational details and configuration-specific descriptions with high-level strategic guidance.


akkki007:
All 13 review comments addressed in cd53752. None disputed; one of them was a substantive factual error on my side and is worth calling out separately.

## The model page finding was correct, and worse than described

The comment on `vision-hypotheticals.md` line 23 said neither cited workflow establishes that model pages are regenerated. Checking it: `publish.yml` line 90 sets `PAGES_DIR="../../meshery.io/collections/_models"` inside a step that iterates `"$PAGES_DIR"/*/*.md` to build OCI tarballs. It reads the tree. It does not write it. And commit `11f9948` shows what `meshery-ci` actually does to a model page:

```
-componentsCount: 27
+- name: arango-platform-link
+  colorIcon: assets/images/integration/kube-arangodb/components/arango-platform-link/...
+componentsCount: 28
```

A surgical in-place edit, not a regeneration. The delete-then-rebuild behavior I cited belongs to `update-catalog.yml`, and it targets `collections/_catalog` only.

Consequences:

- **Hypothetical 1 is demoted from a derived `RESIST` to `Open - awaiting maintainer`.** The repository does not settle whether hand-filled descriptions are acceptable, so it is a maintainer call. The entry now records why the original reasoning was wrong rather than quietly swapping in a different citation that reaches the same conclusion. Split is now 3 derived, 7 open.
- **Principle 1** no longer claims `mesheryctl registry publish website` generates the pages, and now separates registry-data defects from the template, permalink, and rendering defects this repository does own.
- **Principle 2** narrows the destroyed-edit rule to `collections/_catalog` and describes model pages as bot-maintained in place.
- The wrong example in `vision-structure.md` is replaced, with a note on why bundling two trees into one sentence is how a vision ends up asserting more than its evidence supports.

## Remaining findings

| Finding | Change |
| --- | --- |
| `_data/` claim too broad | Narrowed to `_data/discuss/` and `_data/leaderboard.json`. The other 11 entries (`navigation.yml`, `footer.yml`, `features.yml`, and so on) have no producing workflow, and telling contributors not to edit them would have been actively wrong. |
| Collection contract inaccurate | Now matches the config: nine collections with `output`, five with `permalink`, three with `sort_by`. The universal-layout claim is dropped; `collections/_pages` has no layout key. |
| Missing as-of dates on `394` | Added at both sites, per the skill's own rule. |
| Checks do not fail closed | Rewritten. Verifies all three deliverables are readable before scanning, the Python checks `sys.exit(1)`, the hypothetical counts are asserted equal to 10, and the block exits non-zero overall. Verified by hiding a deliverable: previously reported all OK, now exits 1 on the first check. |
| `git add` omits the skill | Fixed. |
| Duplicate `Signed-off-by` | Literal trailer removed; `git commit -s` supplies it. |
| Skill reads another skill's `SKILL.md` | Style rules inlined, with a note explaining why a skill should not consult another skill's instructions. |
| Reference contract assumes maintainer-only verdicts | `hypothetical-bank.md` now documents all three record types with separate templates, plus a rule that a derived verdict whose citation does not hold is demoted to `Open` rather than re-cited. |
| Artifact description in `SKILL.md` | Updated to cover all three verdict types. |

All six conformance checks pass and the block exits 0.

Worth noting for reviewers: the finding above is exactly the class of defect this PR is meant to prevent, and it got through my own verification step because I read a workflow step's variable name instead of the loop body under it. The skill's evidence reference already warns against inferring from filenames; it now also carries the demote-to-`Open` rule that this correction followed.

akkki007:
All four addressed in ab88b48. Two were miscounts on my side that the citations did not actually support.

**docURL overcounted.** Correct: the evidence command matched the `https://docs.meshery.io` host prefix, so it could not support a claim about the `/extensibility/integrations/` path.

```
docURL: https://docs.meshery.io                          -> 394
docURL: https://docs.meshery.io/extensibility/integrations/ -> 377
docURL: https://docs.meshery.io/extensibility/adapters/     ->   8
model pages with no docURL                               ->   0
```

Principle 3 now states the split. (Corrected in place: the adapters figure above read 17, which was 394 minus 377 rather than a measurement. The measured value is 8; the other 9 are installation, task, and guide paths. See the later comment for the full breakdown.) This is a better claim than the one it replaces: every model page defers to docs.meshery.io, under one of two documentation paths, with no exceptions.

**The fifth target was unnamed.** Correct, and it is `meshery/meshery` itself. `publish.yml` commits to five repositories (`./meshery.io` line 126, `./l5` 145, `./meshery` 164, `./meshery-cloud` 183, `./meshery-remote-provider` 202), and the `./meshery` target receives generated model documentation at `docs/content/en/extensions/models` (line 154). That is the source of docs.meshery.io, so the same registry run that fills this site's model pages also fills the documentation those pages link to. Principle 1 names all five now.

**`permalink` does not imply publication.** Correct on both counts. `handbook` declares a permalink but sets `output: false`, and `pages`, `programs`, and `catalog` publish without declaring either key. The wording implied the five permalinks were the published set. Now stated as plain counts (nine `output`, five `permalink`, three `sort_by`) with an explicit note that neither key gates publication. The same phrasing is fixed in the Hypothetical 4 derived verdict.

**Derived verdicts restricted to workflows and config files.** Fair, and the deliverable already contradicted it: `_plugins/catalog_page_metadata.rb` is a Ruby plugin and it is cited as evidence. The definition now reads "a repository file or command: a workflow, a config file, a plugin or script, a committed data file, or the observable result of a command run against the repository," with the file-and-line citation requirement retained.

One note on the research block attached to the fifth-target comment: it states there is no `publish.yml` covering meshery.io, meshery-cloud, and meshery-remote-provider. There is, at `meshery-extensions/integrations-workflow/.github/workflows/publish.yml`, reachable from `meshery/meshery/.github/workflows/integrations-updater.yml` line 25. That does not affect the finding itself, which was right.

All six conformance checks pass, exit 0.

akkki007:
Correct, and fixing it exposed a second error underneath. Both are in 8dcf344.

**The command was not runnable.** The second `grep` had no file operand, so as printed it reads standard input. The 377 was real but not reproducible from what the sheet showed. Both commands now use the same glob and anchor on the front-matter key:

```
grep -l '^docURL: https://docs.meshery.io' collections/_models/*/*.md | wc -l
grep -l '^docURL: https://docs.meshery.io/extensibility/integrations/' collections/_models/*/*.md | wc -l
```

**Re-measuring showed the residual claim was wrong.** The sheet said the remaining 17 point at `/extensibility/adapters/`. Only 8 do:

| Path | Pages |
| --- | --- |
| `/extensibility/integrations/` | 377 |
| `/extensibility/adapters/` | 8 |
| `/installation/...` | 7 |
| `/tasks/performance/` | 1 |
| `/guides/infrastructure-management/` | 1 |

I got the 17 by subtracting 377 from 394 instead of measuring it. Arithmetic is not verification, and in this case the residual hid something worth seeing: two of the installation URLs are malformed, one with a double slash (`/installation//windows`) and one with a trailing space (`/installation/docker `). Those are registry data defects and are visible on the published pages.

`VISION.md` no longer names a single path for the remainder, and the evidence sheet carries the full breakdown.

**Two additions so this class does not recur.** Conformance check 7 flags cited `grep`, `rg`, and `find` commands with no visible file operand. It is advisory rather than failing, because it cannot reliably distinguish a verification command from a workflow snippet quoted as evidence, and a noisy gate gets ignored. It is clean on the current sheet and catches the defect above when reintroduced. The evidence reference gains the two failure modes by name: "the command that will not run" and "the count you subtracted."

Worth noting the first version of that check had a bug of the same family: it split commands on `|` to isolate the pipeline head, which truncates any grep pattern containing alternation and produced two false positives. Fixed before committing.

All seven checks pass, exit 0.

Katotodan (commented):
@akkki007, please scope down this PR to make it easier to review.

I suggest starting with `VISION.md` and opening follow-up PRs for the supporting artifacts and workflow/skill changes. Keeping each PR focused makes it easier for you to validate your changes and allows reviewers to focus on a smaller, well-defined change rather than reviewing a large set of changes at once, which can increase the risk of issues being overlooked.

That said, here is my review of the current `VISION.md`.

Please update it so that it does not become a combination of:

```text
VISION.md
+
REPOSITORY_ARCHITECTURE.md
+
OWNERSHIP.md
+
CONTENT_MODEL.md
+
CONTRIBUTING.md
```

A vision document should primarily communicate the project's **purpose, goals, target users, principles, and non-goals**. The detailed implementation and repository mechanics can be documented separately where necessary.

While working on `VISION.md`, please focus on answering these questions:

* What is the purpose of this repository/project?
* What are its goals?
* Who is it intended to serve?
* What are its non-goals?
* What principles should guide decisions about the project?

You can use the following structure as a starting point:

```md
# Vision

[Short, future-oriented vision]

## Purpose

[Why meshery.io exists]

## Goals

[What meshery.io aims to achieve]

## Principles

### Public web presence
### Present, don't own
### Generated content is generated
### Documentation belongs to docs.meshery.io
### Collections define published content

## Boundaries and Non-Goals

[What meshery.io is not and what it does not aim to become]

## Success Criteria

[How we know the vision is being achieved]
```

I would avoid putting detailed implementation facts such as the exact number of model pages, specific workflow behavior, file-level regeneration mechanics, or individual `_config.yml` settings into `VISION.md`. Those details can change over time and are better suited to repository architecture or contributor documentation.

The goal of `VISION.md` should be to provide a durable **north star for the project and a basis for making decisions**, rather than documenting the current implementation of the repository.

akkki007:
Thank you, this was the right call on both counts.

**Scoped down.** The branch is now a single commit adding `VISION.md` alone, 65 lines, no other files. The evidence sheet, the calibration record, and the skill are preserved on a separate branch and will come as follow-up pull requests once this one lands.

**Rewritten to the suggested structure.** Purpose, Goals, Principles, Boundaries and Non-Goals, Success Criteria, using your five principle headings.

Your diagnosis was accurate. The previous draft really had become an ownership and content-model document wearing a vision title, and the cause was a rule I had set myself: every sentence had to name something specific to this repository so that nothing generic could survive. That rule pushed the document toward page counts, workflow names, and `_config.yml` keys, which is the wrong altitude and, as you say, the part most likely to rot. The current draft has none of those: no counts, no workflow names or behavior, no regeneration mechanics, no config settings.

What I kept is the part that survives implementation change. "Present, don't own" says models, integrations, catalog entries, and community activity originate elsewhere and are published here, so wrong data is fixed at its source while presentation stays this repository's responsibility. That decides where a fix belongs without naming a single workflow.

The mechanics that justified each principle move into the follow-up evidence sheet, which seems like the "repository architecture or contributor documentation" home you described.

**Two things I did not decide on my own.**

I left out any goal about performance, accessibility, or SEO. An earlier draft had one and I removed it because no such check exists in CI, so it would have been a commitment rather than a description. In a north star document an aspirational goal is legitimate, but it is yours to set rather than mine to author. Happy to add one if you want it.

Seven scope questions from the calibration work are still open on #2957 and two of them bear directly on the Boundaries section: whether meshery.io should host full integration documentation rather than linking out, and whether another property may consume this site's rendered pages. I have kept both out of the non-goals list rather than assume an answer. Your input on those would firm up that section.

**Follow-up plan,** in the order you suggested, each separate and small:

1. `docs/vision/vision-evidence.md`, mapping each principle to the repository behavior it was derived from.
2. `.claude/skills/vision/`, the reusable skill, alongside the existing `meshery-blog-writer` skill.

The calibration record folds into the first, or comes as a third, whichever you prefer.

One note on the thread above: the earlier comments respond to automated review of files that are no longer in this pull request. They are left for history rather than as current context.

VISION.md
```diff
@@ -0,0 +1,65 @@
+# Vision
+
+Meshery.io is where people meet Meshery: the place to understand what the project is, browse what it manages, and find a way in.
```
Katotodan: Meshery.io is the digital front door to Meshery, the place to understand what the project is, explore what it manages, and find a way in.

VISION.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Purpose
+
+Meshery.io exists so that anyone evaluating, adopting, or contributing to Meshery can understand the project without reading its source code.
+
+For most people it is the first thing they encounter, and often the only thing they see before deciding whether to go further. The site succeeds when a visitor leaves knowing what Meshery does and where to go next.
```
Katotodan: For most people, it is ...

VISION.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Goals
+
+- Explain what Meshery is and who it is for, in terms a newcomer can act on.
```
Katotodan: Explain what Meshery is and who it is for, in terms a newcomer can understand and act on.

VISION.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Present, don't own
+
+Most of what this site displays belongs to other projects. Models, integrations, catalog entries, and community activity originate elsewhere and are published here. When that data is wrong, it is corrected at its source rather than on the page that displays it.
+
+What meshery.io does own is the presentation: information architecture, layout, navigation, and the reading experience. That distinction decides where a fix belongs.
+
+### Generated content is generated
+
+Content produced by automation stays the responsibility of that automation. Contributors improve the systems that generate it rather than editing the output by hand.
```
Katotodan: Content produced by automation remains the responsibility of the system that generates it. ...

VISION.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Public web presence
+
+Meshery.io is a website, and it is judged as one. A change is good if a visitor understands Meshery better, finds what they came for faster, or leaves with a clearer next step.
+
+### Present, don't own
+
+Most of what this site displays belongs to other projects. Models, integrations, catalog entries, and community activity originate elsewhere and are published here. When that data is wrong, it is corrected at its source rather than on the page that displays it.
+
+What meshery.io does own is the presentation: information architecture, layout, navigation, and the reading experience. That distinction decides where a fix belongs.
```
Katotodan: Use capital letter for mesheri.io "Meshery.io"

Katotodan (commented):
The majority of my comments above are grammatical recommendations; kindly take note of them.

VISION.md
akkki007: Applied, taken verbatim. "Digital front door" is better than what I had, and "explore" reads more naturally than "browse" here.

VISION.md
akkki007: Applied, comma added.

VISION.md
akkki007: Applied. "Understand and act on" is the better pairing; understanding has to come first.

VISION.md
akkki007: Applied. I also reworded the following sentence to "Contributors improve those systems rather than editing their output by hand", since your phrasing made the original repeat "generate it" twice in a row.

VISION.md
akkki007: Applied. The same lowercase form appeared once more on line 41 ("the documentation is canonical and meshery.io links to it"), so I capitalised both rather than only the one flagged.

akkki007:
@Katotodan all five applied in 06eda8c, replied on each thread.

Two small things beyond the literal suggestions:

- The capitalisation fix appeared twice. Line 41 had the same lowercase `meshery.io` in prose, so I corrected both rather than only the flagged one. URLs like `docs.meshery.io` are left lowercase.
- Your rewrite of the generated content line made the next sentence repeat "generate it" twice in a row, so that one now reads "Contributors improve those systems rather than editing their output by hand."

Still one file, 65 lines. Happy to take another pass if anything else reads awkwardly.

VISION.md
```diff
@@ -0,0 +1,65 @@
...
+
+Meshery.io is a website, and it is judged as one. A change is good if a visitor understands Meshery better, finds what they came for faster, or leaves with a clearer next step.
+
+### Present, don't own
+
+Most of what this site displays belongs to other projects. Models, integrations, catalog entries, and community activity originate elsewhere and are published here. When that data is wrong, it is corrected at its source rather than on the page that displays it.
+
+What Meshery.io does own is the presentation: information architecture, layout, navigation, and the reading experience. That distinction decides where a fix belongs.
+
+### Generated content is generated
```
Katotodan: This title could be "Generated content", then the paragraph below describes it.

akkki007: Good catch on the repetition — "Generated content is generated" was saying the same thing twice.

I went with **"Generated content stays generated"** in 3624926, rather than the plain noun. The other headings in the section are all directives — "Present, don't own", "Documentation belongs to docs.meshery.io" — so a bare noun would be the only heading that states a topic instead of a rule, and the rule here is the point: the output is not the place to edit.

If you would still rather have "Generated content", say so and I will switch it — the repetition was the real problem and either version fixes it.

PR #2965, opened 2026-09-10
Fix: Leaderboard nav link redirects out of PR previews

Maanvi212006:
## Problem
Clicking "Leaderboard" in the nav bar on a PR preview deployment sends you out of the preview and onto the live production leaderboard (meshery.io) instead of staying on `/pr-preview/pr-<N>/community/leaderboard/`.

## Root cause
`_data/navigation.yml` points the Leaderboard nav item at `/community/leaderboard` (no trailing slash), but the page's actual permalink is `/community/leaderboard/`. The mismatch means every click first hits the `jekyll-redirect-from`-generated
redirect stub for the no-slash path, which does a client-side `location = "..."` jump built from `site.url` + `site.baseurl`. That extra redirect hop is unnecessary and is the source of the "vanishing" behavior on previews.

This PR fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Leaderboard navigation link to use the correct URL format.


Maanvi212006:
Confirmed the fix is working correctly on the GitHub Pages PR-preview: https://meshery.io/pr-preview/pr-2965/ ; clicking Leaderboard stays on the preview instead of redirecting to production. Closing this out.

Separately, the Netlify deploy-preview (deploy-preview-2965--mesheryio-preview.netlify.app) shows the same symptom, but that's not caused by anything in this codebase; there's no _redirects file or netlify.toml in this repo, so any redirect behavior there is coming from rules configured directly in the Netlify site dashboard (Site configuration → Build & deploy → Redirects and rewrites). That's a separate, infra-level issue and out of scope for this PR. If it needs fixing, it should be tracked separately against the Netlify site config, not the Jekyll source.

PR #2974, opened 2026-09-12
[Blog] Prevent XSS in Search Result Rendering

### LINKED ISSUE #2968: Blog search results render post metadata as raw HTML, allowing XSS
### Description

The blog search implementation builds search result items using `innerHTML` and directly interpolates post metadata such as titles, excerpts, authors, categories, and URLs.

In `js/blog-search.js`, `highlightText()` returns HTML, and `renderResults()` inserts the resulting values into the page without HTML escaping. The main confirmed risk is that **titles, categories, and authors** can be inserted into `innerHTML` without escaping. A malicious or compromised blog post containing HTML or JavaScript in these fields could therefore execute when users search the blog.

`excerpt` and `content` are passed through `strip_html` in `blog/search.json`, making them less directly exploitable through this code path. `result.url` is also interpolated into the generated HTML, but it is generated by Jekyll and is normally repository-controlled.

The repository documentation currently describes the search as having input sanitization, but the existing escaping only handles regex characters used by the search functionality. HTML content in the affected metadata fields is not escaped before being inserted into the DOM.

### Expected Behavior

Blog search results should safely render post metadata, including titles, categories, and authors, without interpreting HTML or JavaScript contained in those values.

User-controlled or repository-provided metadata should not be inserted into `innerHTML` without appropriate HTML escaping. The search results should remain fully functional while safely displaying values containing HTML characters.

### Screenshots

Not applicable. The issue can be demonstrated with a test post containing a value such as:

```html
<img src=x onerror=alert(document.domain)>
```

### Environment:

* OS: macOS
* Browser: Chrome
* Version: Latest
* Device: Desktop
* Page: https://meshery.io/blog/

A possible fix is to construct result elements using DOM APIs and `textContent`, or consistently HTML-escape the affected interpolated values before assigning them to `innerHTML`.

A regression test should verify that HTML-like input in titles, categories, and authors is rendered as text and does not execute.

---

Kunal241207:
**Description**

This PR fixes #2968

Blog search results now render metadata using DOM APIs and `textContent` instead of interpolating values into `innerHTML`. Search highlighting is also constructed safely while preserving the existing result layout and functionality.

**Notes for Reviewers**
- Prevents HTML and JavaScript in titles, categories, authors, and excerpts from being interpreted.
- Replaces unsafe `innerHTML` rendering in blog search results.
- Preserves search highlighting, category links, metadata, and result actions.
- Documentation was updated to reflect the secure rendering behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved blog search highlighting to preserve unmatched text and display matching terms consistently.
  - Search results now render safely, preventing unintended markup from being interpreted as content.
  - Result links now accept only valid HTTP or HTTPS destinations; unsafe or invalid URLs are not linked.

- **Documentation**
  - Updated security documentation to describe safe result rendering and metadata handling.

- **Tests**
  - Added coverage for safe rendering and blocked unsafe link destinations.


Kunal241207:
@banana-three-join @Maanvi212006 @hiyach28 Hello! Could anyone review this PR, Please let me know if anything need changes. Thanks!

akkki007 (commented):
Reviewed this closely because it is a security fix, and ran it locally rather than reviewing by eye alone.

## Test run

```
$ npm install && npm test
added 124 packages in 3s

# tests 2
# pass 2
# fail 0
# duration_ms 14388.29
```

Both tests pass on Node 22.22.0. I also ran `npm run lint`: 94 errors, all of them in `js/calendar.js` and `js/community-popup.js`, which are byte identical to `master` on this branch, so they pre-date this PR. `eslint js/blog-search.js test/blog-search.test.cjs` exits 0, so the files this PR actually touches are clean.

## What I verified

- No HTML sink survives. `grep -n "innerHTML|insertAdjacentHTML|outerHTML|document.write"` over the patched `js/blog-search.js` returns nothing.
- The query is reflected in both summary paths, "No results found for ..." and "Found N results for ...", via `textContent` on a dedicated span rather than string interpolation. That was the spot I most expected to still be vulnerable, and it is handled.
- `result._formatted` is gone from the producer and every consumer, so no render path silently falls back to the old highlighted string.
- Category links stay safe for a different reason worth stating: `slugify` strips everything outside `[\w-]`, and the value is always interpolated into the fixed relative path `/blog/category/<slug>/`, so a protocol can never appear there. That is why `withBaseUrl` rather than `getSafeResultUrl` is correct on that line.
- `javascript:` payloads are blocked. I did not take this on reading; I added a temporary JSDOM case rendering a result with `url: 'javascript:alert(1)'` and asserted on the output:

```
links found: 2
link[0] hasAttribute(href)=false
link[1] hasAttribute(href)=false
```

  It reaches the `URL` protocol allowlist and is rejected, because `withBaseUrl` only short-circuits on `^(?:[a-z]+:)?//` and `javascript:alert(1)` has no `//`.

- The test is a genuine end to end test through JSDOM with real payloads, not a unit test of a helper. `jsdom` is correctly in `devDependencies` and `node --test` avoids adding a framework. Good choices.

## Suggestions, none blocking

Three inline, each verified by running rather than by reading. The scratch tests I used were removed afterwards and are not part of this branch.

This is ready from my side once the author has had a look at them.

test/blog-search.test.cjs
```diff
@@ -0,0 +1,101 @@
...
+  assert.match(resultElement.textContent, /<img src=x onerror=alert\(1\)> Safe title/);
+  assert.match(resultElement.textContent, /<svg onload=alert\(2\)>category/);
+  assert.match(resultElement.textContent, /<script>alert\(3\)<\/script> Author/);
+  assert.match(resultElement.textContent, /<iframe src=javascript:alert\(4\)>Excerpt/);
+
+  dom.window.close();
+});
+
+test('only uses HTTP(S) URLs as result link destinations', async () => {
+  for (const url of ['data:text/html,<script>alert(1)</script>', 'mailto:test@example.com', 'ftp://example.com/file']) {
```
akkki007: Consider adding `javascript:alert(1)` to this list.

It is the canonical vector the allowlist exists to stop, and it takes a different route through the code than the three already here: `data:`, `mailto:` and `ftp:` all reach `getSafeResultUrl` the same way, but `javascript:alert(1)` has no `//`, so it also exercises the `!path.startsWith('/')` early return in `withBaseUrl` before the protocol check runs.

I checked this by running rather than reading. Added a temporary JSDOM case with `url: 'javascript:alert(1)'`:

```
links found: 2
link[0] hasAttribute(href)=false
link[1] hasAttribute(href)=false
```

So it is correctly blocked today. The point of adding it is to keep it blocked: a future change to `withBaseUrl`'s short-circuit conditions could let it through and nothing in the suite would fail.

js/blog-search.js
```diff
@@ -89,10 +93,27 @@ function escapeRegex(string) {
 
 // Simple text highlighting function
 function highlightText(text, query) {
-  if (!text) return '';
+  if (!text) return document.createDocumentFragment();
```
akkki007, line 96: Small defensive nit on `highlightText`.

The guard covers empty `text` but not an empty `query`. With `query = ''`, `escapeRegex` yields `''` and the regex becomes `/()/gi`, which produces a zero-length match at every position.

I reproduced the algorithm standalone to get the exact behaviour rather than estimate it:

```
text="abcde" (5 chars) -> 6 <mark> elements, 6 empty
textContent preserved: true
```

So it is n+1 empty `<mark>` elements for an n character string. A 200 character excerpt would render 201 of them. The visible text survives, so this is DOM bloat rather than corruption.

Not reachable today: the input handler returns early on `query.length < 2`, and `renderResults` is the only caller. So this is about the function being safe on its own rather than a live bug. An extra `if (!query) return document.createTextNode(text);` would make it independent of that caller guard.

js/blog-search.js
```diff
@@ -179,41 +200,80 @@ function renderResults(results, query) {
...
+    }
+
+    const excerptElement = document.createElement('p');
+    excerptElement.appendChild(highlightText(excerpt, query));
+
+    const content = document.createElement('div');
+    content.className = 'post-content';
+    content.appendChild(excerptElement);
+
+    const readMore = document.createElement('a');
```
akkki007: When `getSafeResultUrl` returns null, both the title link and this one are appended with no `href`. Safe, but confirmed inert while still looking interactive. Rendered output from a blocked URL:

```html
<a class="link">Read More</a>
```

An `<a>` without `href` keeps its link styling but is not clickable and not keyboard focusable, so "Read More" becomes text that looks like a control and is not one.

Worth considering rendering a `<span>` instead, or skipping the `button-para` block entirely, so a blocked URL degrades visibly rather than silently. Your test already asserts the no-href behaviour, so whichever way you go it is a one-line test update.

Minor and entirely your call.

Kunal241207:
@akkki007 addressed the suggestions. Thanks for the review.

js/blog-search.js
```diff
@@ -14,6 +14,18 @@ function withBaseUrl(path) {
   return `${siteBaseUrl}${path}`;
 }
 
+function getSafeResultUrl(path) {
+  const url = withBaseUrl(path);
+  if (!url || typeof url !== 'string') return null;
+
+  try {
+    const parsedUrl = new URL(url, window.location.href);
+    return ['http:', 'https:'].includes(parsedUrl.protocol) ? url : null;
```
Maanvi212006: A protocol-relative URL like `//evil.com/x` passes this check too, since `withBaseUrl()` leaves `//` URLs untouched and the browser fills in `https:` for it. so this ends up allowing off-site links, not just safe ones. Might be worth also checking `parsedUrl.origin === window.location.origin` (or requiring the raw value to start with `/`) if the intent is to keep result links on-site.

Kunal241207: @Maanvi212006 Thanks, fixed this and added the test coverage.

Kunal241207:
@Maanvi212006 Hello, I think this is ready for final review. Please take a look when you can.

Kunal241207:
@Maanvi212006 @saurabhraghuvanshii Hi, just a ping in case the pr got missed.

